PR #15, opened 2020-07-14
fixed name in pypi, add badge

xvjiarui:


setup.py
```diff
@@ -77,7 +77,12 @@ def write_version_py():
 def get_version():
     with open(version_file, 'r') as f:
         exec(compile(f.read(), version_file, 'exec'))
-    return locals()['__version__']
+    import sys
+    # return short version for sdist
+    if 'sdist' in sys.argv:
```
hellock, line 82: `if 'sdist' in sys.argv or `bdist_wheel` in sys.argv`

PR #33, opened 2020-07-25
Add load-from flag

nizhib:
The `load-from` option exists both in api and docs, but not in the training script.

This PR fixes that.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -241,11 +241,12 @@ Optional arguments are:
 
 - `--no-validate` (**not suggested**): By default, the codebase will perform evaluation at every k iterations during the training. To disable this behavior, use `--no-validate`.
 - `--work-dir ${WORK_DIR}`: Override the working directory specified in the config file.
-- `--resume-from ${CHECKPOINT_FILE}`: Resume from a previous checkpoint file.
+- `--resume-from ${CHECKPOINT_FILE}`: Resume from a previous checkpoint file (to continue the training process).
+- `--load-from ${CHECKPOINT_FILE}`: Load weights from a checkpoint file (to start finetuning for another task).
 
 Difference between `resume-from` and `load-from`:
-`resume-from` loads both the model weights and optimizer status, and the iteration number is also inherited from the specified checkpoint. It is usually used for resuming the training process that is interrupted accidentally.
-`load-from` only loads the model weights and the training iteration starts from 0. It is usually used for finetuning.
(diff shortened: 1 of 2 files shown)


docs/getting_started.md
```diff
@@ -241,11 +241,12 @@ Optional arguments are:
...
 - `--no-validate` (**not suggested**): By default, the codebase will perform evaluation at every k iterations during the training. To disable this behavior, use `--no-validate`.
 - `--work-dir ${WORK_DIR}`: Override the working directory specified in the config file.
-- `--resume-from ${CHECKPOINT_FILE}`: Resume from a previous checkpoint file.
+- `--resume-from ${CHECKPOINT_FILE}`: Resume from a previous checkpoint file (to continue the training process).
+- `--load-from ${CHECKPOINT_FILE}`: Load weights from a checkpoint file (to start finetuning for another task).
 
 Difference between `resume-from` and `load-from`:
-`resume-from` loads both the model weights and optimizer status, and the iteration number is also inherited from the specified checkpoint. It is usually used for resuming the training process that is interrupted accidentally.
-`load-from` only loads the model weights and the training iteration starts from 0. It is usually used for finetuning.
+- `resume-from` loads both the model weights and optimizer state including the iteration number.
```
xvjiarui, line 248: We may keep `resume-from` part unchanged.

docs/getting_started.md
```diff
@@ -241,11 +241,12 @@ Optional arguments are:
...
+- `load-from` loads only the model weights, starts the training from scratch (iteration 0).
```
xvjiarui, line 249:
```suggestion
- `load-from` loads only the model weights, starts the training from scratch (iteration 0). It is usually used for fine-tuning. 
```

xvjiarui, line 249: From scratch could be confusing. We may take the original description.

xvjiarui:
Hi @nizhib 
Thanks for the fix. Please revise according to the comments. 
If you find something wrong in the original description, please let me know.

nizhib:
The idea was to leave "what scenario is it for" in the initial args description and have not so long "what technically happens" in the second one. Is it enough to change "from scratch (iteration 0)" -> "from iteration 0" to avoid confusion and maybe expand the `resume-from` part with "loads both the model weights and optimizer state including the iteration number_, continues the training" for `resume-from` or should I revert the changes for the latter?

xvjiarui:
Hi @nizhib 
Thanks for the explanation. I got your point. 
I think both "what scenario is it for" and "what technically happens" may be useful for users.

PR #35, opened 2020-07-26
Add Semantic FPN and PointRend

xvjiarui:


baibaidj:
May I ask when PointRend will be released?

hellock:
Split into #94 and #109.

PR #47, opened 2020-08-04
Support ResNeSt backbone

xvjiarui:


mmseg/models/backbones/resnest.py
```diff
@@ -0,0 +1,310 @@
...
+        if self.radix > 1:
+            x = x.view(batch, self.groups, self.radix, -1).transpose(1, 2)
+            x = F.softmax(x, dim=1)
+            x = x.reshape(batch, -1)
+        else:
+            x = torch.sigmoid(x)
+        return x
+
+
+class SplAtConv2d(nn.Module):
```
hellock, line 38: `SplitAttentionConv2d`

mmseg/models/backbones/resnest.py
```diff
@@ -0,0 +1,310 @@
...
+from ..utils import ResLayer
+from .resnet import Bottleneck as _Bottleneck
+from .resnet import ResNetV1d
+
+
+class RSoftmax(nn.Module):
+    """Radix Softmax module in ``SplAtConv2d``.
+
+    Args:
+        radix (int): Radix of input. groups (int): Groups of input.
```
hellock, line 19: Note the formating.

mmseg/models/backbones/resnest.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def __init__(self, radix, groups):
+        super().__init__()
+        self.radix = radix
+        self.groups = groups
+
+    def forward(self, x):
+        batch = x.size(0)
+        if self.radix > 1:
+            x = x.view(batch, self.groups, self.radix, -1).transpose(1, 2)
```
hellock, line 30: Is the `transpose` necessary? Can we specify `dim=2` for softmax?

mmseg/models/backbones/resnest.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def forward(self, x):
+        x = self.conv(x)
+        x = self.norm0(x)
+        x = self.relu(x)
+
+        batch, rchannel = x.shape[:2]
+        if self.radix > 1:
+            splited = torch.split(x, rchannel // self.radix, dim=1)
+            gap = sum(splited)
```
hellock, line 124: Use `torch.sum(splited, dim=0)` instead.

mmseg/models/backbones/resnest.py
```diff
@@ -0,0 +1,310 @@
...
+        return getattr(self, self.norm1_name)
+
+    def forward(self, x):
+        x = self.conv(x)
+        x = self.norm0(x)
+        x = self.relu(x)
+
+        batch, rchannel = x.shape[:2]
+        if self.radix > 1:
+            splited = torch.split(x, rchannel // self.radix, dim=1)
```
hellock, line 123: The passive voice of `split` is `split`, not `splited`.

mmseg/models/backbones/resnest.py
xvjiarui: `splited` is a list, `torch.sum` doesn't support.

mmseg/models/backbones/resnest.py
xvjiarui: `x` is reshaped to original shape afterward, which means the stack order of `self.group` and `self.radix` is changed (although `self.group=1` in current config).

mmseg/models/backbones/resnest.py
hellock:
```python
x = x.view(batch, rchannel // self.radix, -1, x.shape[2], x.shape[3])
gap = x.sum(dim=2)
```
may be faster?

xvjiarui: Nice.

PR #55, opened 2020-08-07
Auto get version info and git hash

xvjiarui:


hellock:
We may bump version to 0.5.1

hellock:
Update the doc:
image

docs/install.md
```diff
@@ -54,10 +54,9 @@ pip install -e .  # or "python setup.py develop"
 
 Note:
 
-1. In `dev` mode, the git commit id will be written to the version number with step *d*, e.g. 0.5.0+c415a2e. The version will also be saved in trained models.
-It is recommended that you run step *d* each time you pull some updates from github. If C++/CUDA codes are modified, then this step is compulsory.
+1. The git commit id will be written to the version number with step *d*, e.g. 0.5.0+c415a2e. The version will also be saved in trained models.
```
hellock, line 57: This is wrong. The version number does not contain the git hash.

mmseg/version.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) Open-MMLab. All rights reserved.
+
+__version__ = '0.5.1'
```
hellock, line 3: Add changelog.

hellock:
ping @xvjiarui

PR #71, opened 2020-08-18
Add support for custom classes

igonro:
Please, read https://github.com/open-mmlab/mmdetection/pull/2408 for understanding the changes and the justification.

I didn't add the boolean parameter `custom_classes` (as it is currently in **mmdetection**), because I didn't know if it would fit or if it would be necessary to do it. Everything else has been done with the intention of following the logic as it is in **mmdetection**.


xvjiarui:
Hi @igonro 
Thanks for your contribution. 
But in this PR, training on the only subclass is still not supported since the annotation is the same. 
We need to modify data pipeline to make it work.

igonro:
Is that functionality already being implemented? If not, I could try to implement it. @xvjiarui

xvjiarui:
Hi @igonro 
It not implemented yet. PRs are welcome!

igonro:
Have you identified any problem or any way to proceed to implement it? I'm still a newbie to mmsegmentation, so it would be great to get some hint (if you have thought about it). @xvjiarui

igonro:
Hi @xvjiarui, can you check out the last changes I have made?

mmseg/datasets/custom.py
```diff
@@ -85,6 +86,32 @@ def __init__(self,
...
+                                self.reduce_zero_label)
+                    else:
+                        self.old_id_to_new_id[i + int(
+                            self.reduce_zero_label)] = classes.index(c) + int(
+                                self.reduce_zero_label)
+                        matched_classes += 1
+                if matched_classes < len(classes):
+                    raise ValueError('classes is not a subset of CLASSES.')
+
+            self.CLASSES = classes
```
xvjiarui, line 114: Could we do something like [this](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/custom.py#L231-L254)?

xvjiarui, line 114: The `PALETTE` should be modified accordingly.

mmseg/datasets/custom.py
```diff
@@ -160,6 +187,8 @@ def get_ann_info(self, idx):
     def pre_pipeline(self, results):
         """Prepare results dict for pipeline."""
         results['seg_fields'] = []
+        if self.custom_classes:
+            results['old_id_to_new_id'] = self.old_id_to_new_id
```
xvjiarui, line 191: Could we use something like `label_map` and add some comments about it?

mmseg/datasets/custom.py
```diff
@@ -85,6 +86,32 @@ def __init__(self,
...
+                    f'Unsupported type {type(classes)} of classes.')
+
+            if self.CLASSES:
+                self.old_id_to_new_id = {}
+                matched_classes = 0
+                for i, c in enumerate(self.CLASSES):
+                    if c not in classes:
+                        self.old_id_to_new_id[
+                            i + int(self.reduce_zero_label)] = -1 + int(
+                                self.reduce_zero_label)
```
xvjiarui, line 105: `reduce_zero_label=True` for a dataset, the CLASSES is already reduced, like ADE20k.

mmseg/datasets/custom.py
```diff
@@ -85,6 +86,32 @@ def __init__(self,
...
+                    else:
+                        self.old_id_to_new_id[i + int(
+                            self.reduce_zero_label)] = classes.index(c) + int(
+                                self.reduce_zero_label)
+                        matched_classes += 1
+                if matched_classes < len(classes):
+                    raise ValueError('classes is not a subset of CLASSES.')
```
xvjiarui, line 112: We may check this by python builtin `issubset`.

mmseg/datasets/pipelines/loading.py
```diff
@@ -132,6 +132,10 @@ def __call__(self, results):
         gt_semantic_seg = mmcv.imfrombytes(
             img_bytes, flag='unchanged',
             backend=self.imdecode_backend).squeeze().astype(np.uint8)
+        # modify if custom classes
+        if results.get('old_id_to_new_id', None) is not None:
+            for old_id, new_id in results['old_id_to_new_id'].items():
+                gt_semantic_seg[gt_semantic_seg == old_id] = new_id
```
xvjiarui, line 138: We need to do something similar in `CustomDataset.get_gt_seg_maps()`.

mmseg/datasets/custom.py
```diff
@@ -85,6 +86,32 @@ def __init__(self,
...
+            self.custom_classes = True
+            if not isinstance(classes, (tuple, list)):
+                raise ValueError(
+                    f'Unsupported type {type(classes)} of classes.')
+
+            if self.CLASSES:
+                self.old_id_to_new_id = {}
+                matched_classes = 0
+                for i, c in enumerate(self.CLASSES):
+                    if c not in classes:
```
xvjiarui, line 102: We may remove this condition by checking `issubset` first.

mmseg/datasets/custom.py
igonro: I don't understand what you mean by `label_map`. Do you mean changing the key name?

mmseg/datasets/custom.py
igonro: If I understood correctly, you suggest using `CLASSES` for the labels, and then transforming the background to -1 in LoadAnnotation using `reduce_zero_label` flag. Is this correct?

mmseg/datasets/custom.py
xvjiarui: Sorry for the ambiguity. 
Yes, `'old_id_to_new_id'` is a bit too long. But `'label_map'` may not be straight forward enough. Any idea?

mmseg/datasets/custom.py
xvjiarui: Yes. We only need to handle `reduce_zero_label` in `LoadAnnotation` and `CustomDataset.get_gt_seg_maps()`.

mmseg/datasets/custom.py
```diff
@@ -230,6 +238,66 @@ def get_gt_seg_maps(self):
...
+
+        else:
+            palette_values = []
+            for x in sorted(self.old_id_to_new_id.items(), key=lambda x: x[1]):
+                if x[1] != -1:
+                    palette_values.append(self.PALETTE[x[0]])
+
+            palette_values = type(self.PALETTE)(palette_values)
+
+        return palette_values
```
xvjiarui, line 299: Custom palette from user input may not be necessary for `CustomDataset`. We may directly set `palette` by indices from subclasses.

igonro: Yes, sorry. I was still fixing some things... You can now take a look if you can.

mmseg/datasets/custom.py
igonro: Maybe `class_transform`, `class_id_transform` or `label_transform`?

igonro: Or `label_map` could be good with a comment explaining it.

xvjiarui: Then `label_map` it is. We may add comments about key and value of this dict.

xvjiarui: We may also add comments about what's the purpose of adding this dict.

tests/test_data/test_dataset.py
```diff
@@ -171,3 +172,62 @@ def test_custom_dataset():
...
+
+    # Test default behavior
+    custom_dataset = dataset_class(
+        pipeline=[],
+        img_dir=MagicMock(),
+        split=MagicMock(),
+        classes=None,
+        test_mode=True)
+
+    assert custom_dataset.CLASSES == original_classes
```
xvjiarui, line 233: We may add a test case for `LoadAnnotation` pipeline when there is custom classes. 
To achieve this, we may generate a random ground truth segmentation map.

xvjiarui:
You may change the draft status when this PR is ready.

tests/test_data/test_dataset.py
igonro: Should I add it in `test_loading.py` or in this same file?

mmseg/datasets/custom.py
```diff
@@ -230,6 +244,66 @@ def get_gt_seg_maps(self):
...
+                    self.old_id_to_new_id[i] = classes.index(c)
+
+        return class_names, self.get_palette_for_custom_classes(palette)
+
+    def get_palette_for_custom_classes(self, palette=None):
+
+        if palette:
+            if isinstance(palette, str):
+                # take it as a file path
+                palette = mmcv.list_from_file(palette)
```
xvjiarui, line 293: The `palette` is a list of list. It is not trivial to load palette from path. 
I suggest removing this function. We just need to select the right `palette` if there is.

mmseg/datasets/custom.py
xvjiarui: So shall we rename it to `label_map`?

tests/test_data/test_dataset.py
xvjiarui: We may add it in `test_loading.py`. 
We should check if the loaded annotation is of custom classes as desired.

tests/test_data/test_loading.py
```diff
@@ -98,3 +100,99 @@ def test_load_seg(self):
...
+    def test_load_seg_custom_classes(self):
+
+        test_img = np.random.rand(10, 10)
+        test_gt = np.zeros_like(test_img)
+        test_gt[2:4, 2:4] = 1
+        test_gt[2:4, 6:8] = 2
+        test_gt[6:8, 2:4] = 3
+        test_gt[6:8, 6:8] = 4
+
+        img_path = osp.join(osp.dirname(__file__), 'img.jpg')
```
igonro, line 113: I've tried using `tmpdir` and `tmp_path` for doing this, but it raises an error when calling `imwrite` (not sure why), so I've had to make this workaround.

igonro:
Hey @xvjiarui, could you review this when you have the time? Thanks!

xvjiarui:
Hi @igonro 
Sorry for the late reply. 
[This](https://github.com/open-mmlab/mmsegmentation/pull/71#discussion_r484018278) comment has not be resolved yet. 
Could you resolve it first? I will look into [`tmpdir` issue](https://github.com/open-mmlab/mmsegmentation/pull/71#discussion_r484378561) today.

igonro:
> Hi @igonro
> Sorry for the late reply.
> [This](https://github.com/open-mmlab/mmsegmentation/pull/71#discussion_r484018278) comment has not be resolved yet.
> Could you resolve it first? I will look into [`tmpdir` issue](https://github.com/open-mmlab/mmsegmentation/pull/71#discussion_r484378561) today.

Fixed. I've missunderstood your comment.

mmseg/datasets/custom.py
```diff
@@ -74,7 +78,9 @@ def __init__(self,
                  data_root=None,
                  test_mode=False,
                  ignore_index=255,
-                 reduce_zero_label=False):
+                 reduce_zero_label=False,
+                 classes=None,
+                 palette=None):
```
xvjiarui, line 83: `palette` arg is not needed.

mmseg/datasets/custom.py
```diff
@@ -230,6 +244,49 @@ def get_gt_seg_maps(self):
 
         return gt_seg_maps
 
+    def get_classes_and_palette(self, classes=None, palette=None):
```
xvjiarui, line 247: `palette` arg is not needed.

mmseg/datasets/custom.py
```diff
@@ -230,6 +244,49 @@ def get_gt_seg_maps(self):
...
+            # are the new label ids.
+            # used for changing pixel labels in load_annotations.
+            self.label_map = {}
+            for i, c in enumerate(self.CLASSES):
+                if c not in class_names:
+                    self.label_map[i] = -1
+                else:
+                    self.label_map[i] = classes.index(c)
+
+        return class_names, palette
```
xvjiarui, line 288: We need to return the subset of `PALETTE` since we are using a subset of `CLASSES`.

xvjiarui:
> > Hi @igonro
> > Sorry for the late reply.
> > [This](https://github.com/open-mmlab/mmsegmentation/pull/71#discussion_r484018278) comment has not be resolved yet.
> > Could you resolve it first? I will look into [`tmpdir` issue](https://github.com/open-mmlab/mmsegmentation/pull/71#discussion_r484378561) today.
> 
> Fixed. I've missunderstood your comment.

Hi @igonro 
There are still some uncleaned snippets about `palette`. Please refer to the comment for detail.

igonro:
Hi @xvjiarui 
I made those changes. I hope that now everything is fine.

xvjiarui (approved):

PR #93, opened 2020-08-25
Update mmseg/apis/test.py to fix the problem of if statement.

### LINKED ISSUE #92: A mistake
https://github.com/open-mmlab/mmsegmentation/blob/bafc0e5db679b66a739db626dd57a48f8d3a4f6e/mmseg/apis/test.py#L34

It should be:

```python
        if isinstance(result, list):
            results.extend(result)
        else:
            results.append(result)
```

In this case, the if statement here is reasonable.

lartpang:
Fixed #92

### DIFF
--- a/mmseg/apis/test.py
+++ b/mmseg/apis/test.py
@@ -31,7 +31,7 @@ def single_gpu_test(model, data_loader, show=False, out_dir=None):
     for i, data in enumerate(data_loader):
         with torch.no_grad():
             result = model(return_loss=False, **data)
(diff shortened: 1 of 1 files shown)


xvjiarui:
Hi @lartpang 
Thanks for your contribution. 
The code looks good to me. 
Could you please sign CLA?

lartpang:
@xvjiarui I have signed it.

xvjiarui (approved):

PR #94, opened 2020-08-26
Add Semantic FPN

xvjiarui:


hellock:
This PR contains both Semantic FPN and HRFPN. Better to raise two PRs.

hellock (approved):

PR #98, opened 2020-08-27
Mobilenet V3 and fastscnn develop WIP

johnzja:
(i) This PR fixes problems of inverted residual module in fast-scnn. inverted residual module is moved to mmseg/models/utils, and is kept even with mmcls.
(ii) Some problems may still exist in mobilenet v3. The latest model is still training.
(iii) Unit tests are needed for MobileNet v3.


xvjiarui:
We may separate MobileNetV3 and Fast-SCNN fix into 2 PRs.

johnzja:
> We may separate MobileNetV3 and Fast-SCNN fix into 2 PRs.

Problems in Fast-SCNN do not affect correct training & Testing. You can merge this PR until MobileNetV3 is complete.

mmseg/models/utils/inverted_residual.py
```diff
@@ -0,0 +1,120 @@
+import torch.nn as nn
+import torch.utils.checkpoint as cp
+from mmcv.cnn import ConvModule
+
+from .se_layer import SELayer
+
+
+class InvertedResidual(nn.Module):
```
xvjiarui, line 8: There is also one `InvertedResidual` in `mmseg/model/utils/`. 
We should have a consistent one.

johnzja: I've merged these two "inverted residual" files. Now the final version of "InvertedResidual" is in mmseg/model/utils only.

PR #116, opened 2020-09-09
fix acc and iou compute nan problem

sshuair:
Reproduce
```python
In [4]: import torch
   ...: targets = torch.randint(0, 5, size=(12,3,256,256))
   ...: preds = torch.randint(0,4, size=(12,3,256,256))
   ...: mean_iou(preds, targets,7, ignore_index=255)
/usr/local/bin/ipython:63: RuntimeWarning: invalid value encountered in true_divide
[471730. 473621. 472044. 471450. 470451.      0.      0.]
/usr/local/bin/ipython:65: RuntimeWarning: invalid value encountered in true_divide

Out[4]:
(0.20025168524848092,
 array([0.25013461, 0.25040275, 0.25065036, 0.2493223 , 0.     ,    nan,        nan]),
 array([0.12510297, 0.12543045, 0.12530315, 0.1246835 , 0.        ,    nan,        nan]))
```

new result:

```python
In [6]: import torch
   ...: targets = torch.randint(0, 5, size=(12,3,256,256))
   ...: preds = torch.randint(0,4, size=(12,3,256,256))
   ...: mean_iou(preds, targets,7, ignore_index=255)
[472809. 472306. 470679. 471310. 472192.      0.      0.]
Out[6]:
(0.19984902275933158,
 array([0.24924229, 0.25092842, 0.25056355, 0.2486877 , 0.      , 0.        , 0.        ]),
 array([0.12472205, 0.12560796, 0.12515188, 0.12413867, 0.        ,   0.        , 0.        ]))

In [7]:
```

### DIFF
--- a/mmseg/core/evaluation/mean_iou.py
+++ b/mmseg/core/evaluation/mean_iou.py
@@ -34,7 +34,7 @@ def intersect_and_union(pred_label, label, num_classes, ignore_index):
     return area_intersect, area_union, area_pred_label, area_label
 
 
-def mean_iou(results, gt_seg_maps, num_classes, ignore_index):
+def mean_iou(results, gt_seg_maps, num_classes, ignore_index, nan_to_num=None):
     """Calculate Intersection and Union (IoU)
 
     Args:
@@ -42,6 +42,7 @@ def mean_iou(results, gt_seg_maps, num_classes, ignore_index):
         gt_seg_maps (list[ndarray]): list of ground truth segmentation maps
         num_classes (int): Number of categories
         ignore_index (int): Index that will be ignored in evaluation.
+        nan_to_num (int): Replace NaN with the numbers defined by the user.
 
      Returns:
          float: Overall accuracy on all images.
@@ -64,7 +65,8 @@ def mean_iou(results, gt_seg_maps, num_classes, ignore_index):
         total_area_pred_label += area_pred_label
         total_area_label += area_label
     all_acc = total_area_intersect.sum() / total_area_label.sum()
-    acc = total_area_intersect / (total_area_label + 2.220446049250313e-16)
-    iou = total_area_intersect / (total_area_union + 2.220446049250313e-16)
-
+    acc = total_area_intersect / total_area_label
+    iou = total_area_intersect / total_area_union
+    if nan_to_num is not None:
+        return all_acc, np.nan_to_num(acc, nan=nan_to_num), np.nan_to_num(iou, nan=nan_to_num)
     return all_acc, acc, iou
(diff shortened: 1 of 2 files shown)


xvjiarui:
Hi @sshuair 
Thanks for your contribution. 
May I know the motivation for this PR? Since NaN and 0 in mIoU have different meanings. We may just keep the NaN as it is.

sshuair:
@xvjiarui 🤔 There is a problem with NAN display in TensorBoard.  Maybe I should handle it when write out to tensorboard.

xvjiarui:
Hi @sshuair 
Sorry for the late reply.
How about this? Setting nan to -1 by [this](https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html).

xvjiarui:
Btw, could you please sign the CLA?

sshuair:
There is a problem with -1. 
If we use it in the loop, the -1 value will lead error result.

```python
iou = []
for inputs, targets in dataloader:
    outputs = model(inputs)
    _, accs, ious = mean_iou(preds, targets,7, ignore_index=255)
    iou.append(ious)

# calcuate the mean iou of the evalidation datasets
print(np.mean(iou, 1))
```
If some iter is -1 , the whole evalidation datasets result will getting lower than the true result.

for example:

- iter1 = [0.2501346 , 0.2493223   ,  -1,                       -1                ] # not contain class 2 and class 3
- iter2 = [0.2501346 , 0.2493223  ,    0.2493223,        0.2493223]

the average will lower than 0, obviously this calculation is not correct.

xvjiarui:
@sshuair 
I got your point. Could we add an argument called `nan_to_num=None`. By default, we don't do anything. If `nan_to_num` is specified, we could just set `nan` by it.

sshuair:
I agree with that, I will submit new code.

mmseg/core/evaluation/mean_iou.py
```diff
@@ -34,14 +34,15 @@ def intersect_and_union(pred_label, label, num_classes, ignore_index):
...
-def mean_iou(results, gt_seg_maps, num_classes, ignore_index):
+def mean_iou(results, gt_seg_maps, num_classes, ignore_index, nan_to_num=None):
     """Calculate Intersection and Union (IoU)
 
     Args:
         results (list[ndarray]): List of prediction segmentation maps
         gt_seg_maps (list[ndarray]): list of ground truth segmentation maps
         num_classes (int): Number of categories
         ignore_index (int): Index that will be ignored in evaluation.
+        nan_to_num (int): Replace NaN with the numbers defined by the user.
```
xvjiarui, line 45:
```suggestion
        nan_to_num (int, optional): If specified, NaN values will be replaced by the numbers defined by the user. Default: None. 
```

xvjiarui (approved):

PR #118, opened 2020-09-09
Add runner type

daavoo:
This P.R. uses mmcv's `build_runner` in order to  allow to select between EpochBasedRunner and IterBasedRunner. Defaults to IterBasedRunner in order to maintain the original behaviour.


mmseg/apis/train.py
```diff
@@ -70,7 +71,14 @@ def train_segmentor(model,
     # build runner
     optimizer = build_optimizer(model, cfg.optimizer)
 
-    runner = IterBasedRunner(
+    if runner_type == 'iter':
```
xvjiarui, line 74: By default, all runners are `IterBaseRunner`. 
We may use `IterBaseRunner` if no `runner_type` is not specified.

tools/train.py
```diff
@@ -53,6 +53,11 @@ def parse_args():
         default='none',
         help='job launcher')
     parser.add_argument('--local_rank', type=int, default=0)
+    parser.add_argument(
+        '--runner-type',
+        choices=['epoch', 'iter'],
+        default='iter',
+        help='Type of mmcv runner to use. EpochBased or IterBased.')
```
xvjiarui, line 60: In the current implementation, we use args to determine which runner to use. 
However, some config fields are also related to the runner type. We may try to choose runner type from config file. So that users could directly use EpochRunner by changing config. 
For example, if there is no `runner_type` in the config, `IterRunner` will be used. Otherwise, `EpochRunner` will be used.

daavoo: Following that path, maybe it's a good idea to add a "build_runner" using Registry to `mmcv`, so the configuration is similar to the "optimizer" or other fields that use the registry?? 

It would also allow the users to easily try custom runners

xvjiarui: Yep. Sounds good. And we could set the default runner to be `IterRunner` for MMSeg. 
You may create a PR in MMCV to add a runner registry.

daavoo: Will do. Making this a draft untill mmcv is updated

daavoo:
This P.R. needs 1.1.4 which is not available for installation in C.I.

mmseg/apis/train.py
```diff
@@ -70,13 +70,15 @@ def train_segmentor(model,
...
 
-    runner = IterBasedRunner(
-        model=model,
-        batch_processor=None,
-        optimizer=optimizer,
-        work_dir=cfg.work_dir,
-        logger=logger,
-        meta=meta)
+    runner = build_runner(
+        cfg.runner,
```
hellock, line 74: This breaks the backward compatibility. It should be better to keep the original api and add a deprecation warning.

daavoo: I was not sure about how to maintain  the backward compatibility so I create a default runner section that reproduces the old behavior.

mmseg/apis/train.py
```diff
@@ -70,13 +71,21 @@ def train_segmentor(model,
...
-        batch_processor=None,
-        optimizer=optimizer,
-        work_dir=cfg.work_dir,
-        logger=logger,
-        meta=meta)
+    if cfg.get('runner') is None:
+        cfg.runner = {'type': 'IterBasedRunner', 'max_iters': cfg.total_iters}
+        warnings.warn(
+            'config is now expected to have a `runner` section, '
+            'please set `runner` in your config.', DeprecationWarning)
```
xvjiarui, line 78: `DeprecationWarning` will not be shown by default. Just `UserWarning` is fine.

mmseg/apis/train.py
xvjiarui: Currently, if the user is using an old config file,  there will be error since `max_iters` is not specified.

xvjiarui: We may keep the backward compatibility so that user may use old config file.

daavoo: With the block added above:

```python

    if cfg.get('runner') is None:
        cfg.runner = {'type': 'IterBasedRunner', 'max_iters': cfg.total_iters}
        warnings.warn(
            'config is now expected to have a `runner` section, '
            'please set `runner` in your config.', DeprecationWarning)
```
User should be able to use an old config file, right?

hellock:
https://github.com/open-mmlab/mmclassification/pull/54

hellock:
LGTM. @xvjiarui may take a pass.

xvjiarui (approved):

PR #152, opened 2020-09-24
Fix cpu inference

daavoo:
Add missing map_location

### DIFF
--- a/mmseg/apis/inference.py
+++ b/mmseg/apis/inference.py
@@ -28,7 +28,8 @@ def init_segmentor(config, checkpoint=None, device='cuda:0'):
     config.model.pretrained = None
     model = build_segmentor(config.model, test_cfg=config.test_cfg)
     if checkpoint is not None:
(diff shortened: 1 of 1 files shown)


mmseg/apis/inference.py
```diff
@@ -28,7 +29,8 @@ def init_segmentor(config, checkpoint=None, device='cuda:0'):
     config.model.pretrained = None
     model = build_segmentor(config.model, test_cfg=config.test_cfg)
     if checkpoint is not None:
-        checkpoint = load_checkpoint(model, checkpoint)
+        map_loc = 'cpu' if device == 'cpu' else None
+        checkpoint = load_checkpoint(model, checkpoint, map_location=map_loc)
```
xvjiarui, line 33: Initially `model` is on `cpu`. 
So just 
```suggestion
        checkpoint = load_checkpoint(model, checkpoint, map_location='cpu')
```
should do the trick.

xvjiarui (approved):

PR #157, opened 2020-09-25
Support custom palette

daavoo:
`CustomDataset` currently supports setting `classes` via config file, covering two use cases: using your own dataset and using a subset of existing dataset. However because of how it's currently handled, the palette only covers the subset case. 

This P.R. includes the option to set `palette` via configuration files and automatically creates one with random colors when no palette is given.

### DIFF
--- a/tests/test_data/test_dataset.py
+++ b/tests/test_data/test_dataset.py
@@ -231,3 +231,33 @@ def test_custom_classes_override_default(dataset, classes):
         test_mode=True)
 
     assert custom_dataset.CLASSES == original_classes
+
+
+@patch('mmseg.datasets.CustomDataset.load_annotations', MagicMock)
+@patch('mmseg.datasets.CustomDataset.__getitem__',
+       MagicMock(side_effect=lambda idx: idx))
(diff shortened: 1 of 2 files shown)


hellock:
ping @xvjiarui

tests/test_data/test_dataset.py
```diff
@@ -231,3 +231,33 @@ def test_custom_classes_override_default(dataset, classes):
...
+    assert len(dataset.PALETTE) == 2
+    for class_color in dataset.PALETTE:
+        assert len(class_color) == 3
+        assert all(x >= 0 and x <= 255 for x in class_color)
+
+
+@patch('mmseg.datasets.CustomDataset.load_annotations', MagicMock)
+@patch('mmseg.datasets.CustomDataset.__getitem__',
+       MagicMock(side_effect=lambda idx: idx))
+def test_custom_dataset_rcustom_palette():
```
xvjiarui, line 255: May be ```suggestion
def test_custom_dataset_custom_palette():
```

xvjiarui (approved):

PR #227, opened 2020-11-04
[Feature] Add Rgb2Gray transform

yamengxi:


mmseg/datasets/pipelines/transforms.py
```diff
@@ -208,6 +208,8 @@ def __call__(self, results):
                 'keep_ratio' keys are added into result dict.
         """
 
+        # print(results)
+        print(results, results['img'].shape)
```
xvjiarui, line 212: Delete these.

mmseg/datasets/pipelines/transforms.py
```diff
@@ -463,6 +465,61 @@ def __repr__(self):
         return self.__class__.__name__ + f'(crop_size={self.crop_size})'
 
 
+@PIPELINES.register_module()
+class Rgb2Gray(object):
+    """Convert RGB image to grayscale image.
+
+    This transform calculate the weighted mean of input image channels with
+    ``weights`` and then expand the channels to ``out_channels``. When
+    ``out_channels`` equals 0, the number of output channels is the same as
```
xvjiarui, line 474: Use `None` instead of 0.

mmseg/datasets/pipelines/transforms.py
```diff
@@ -463,6 +465,61 @@ def __repr__(self):
         return self.__class__.__name__ + f'(crop_size={self.crop_size})'
 
 
+@PIPELINES.register_module()
+class Rgb2Gray(object):
```
xvjiarui, line 469: RGB2Gray

mmseg/datasets/pipelines/transforms.py
```diff
@@ -463,6 +465,61 @@ def __repr__(self):
...
+    input channels.
+
+    Args:
+        out_channels (int): Expected number of output channels after
+            transforming.
+        weights (tuple[float]): The weights to calculate the weighted mean.
+            Default: (0.299, 0.587, 0.114).
+    """
+
+    def __init__(self, out_channels, weights=(0.299, 0.587, 0.114)):
```
xvjiarui, line 484: Why using weighted sum instead of direct `mmcv.rgb2gray`.

yamengxi: By this method, we can use the weights defined by us instead of fixed weights.

mmseg/datasets/pipelines/transforms.py
```diff
@@ -463,6 +465,61 @@ def __repr__(self):
...
+            results (dict): Result dict from loading pipeline.
+
+        Returns:
+            dict: Result dict with grayscale image.
+        """
+        img = results['img']
+        assert len(img.shape) == 3
+        assert img.shape[2] == len(self.weights)
+        weights = np.array(self.weights).reshape((1, 1, -1))
+        img = (img * weights).sum(2)
```
xvjiarui, line 505: We may use `keep_dim=True`. In this way, we don't need to expand dim.

hellock:
Task linked: [CU-fjrgpn Data preprocessing](https://app.clickup.com/t/fjrgpn)

xvjiarui (approved):

PR #240, opened 2020-11-11
Support image based class weighted CE

Junjun2016:
Support image based class weighted CE

### DIFF
--- a/mmseg/models/losses/cross_entropy_loss.py
+++ b/mmseg/models/losses/cross_entropy_loss.py
@@ -6,9 +6,22 @@ from ..builder import LOSSES
 from .utils import weight_reduce_loss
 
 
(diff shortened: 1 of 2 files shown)


mmseg/models/losses/cross_entropy_loss.py
```diff
@@ -6,22 +6,58 @@
...
+    else:
+        weights = ((hist != 0) * upper_bound * (1 - hist_norm)) + 1
+    return weights
+
+
 def cross_entropy(pred,
                   label,
                   weight=None,
+                  img_based_class_weights=None,
+                  batch_weights=True,
```
xvjiarui, line 24: I suggest making `CrossEntropy` simple as it is. 
We may add a new loss module.

PR #247, opened 2020-11-11
Add relaxed boundary one hot transform

Junjun2016:
Add relaxed boundary one hot transform

### DIFF
--- a/mmseg/datasets/pipelines/transforms.py
+++ b/mmseg/datasets/pipelines/transforms.py
@@ -2,6 +2,7 @@ import mmcv
 import numpy as np
 from mmcv.utils import deprecated_api_warning
 from numpy import random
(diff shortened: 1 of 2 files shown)


mmseg/datasets/pipelines/transforms.py
```diff
@@ -680,6 +681,83 @@ def __repr__(self):
         return self.__class__.__name__ + f'(scale_factor={self.scale_factor})'
 
 
+@PIPELINES.register_module()
+class BoundaryRelaxedOneHot(object):
+    """Convert segmentation label to one hot with boundary ralax.
```
xvjiarui, line 686: ralax -> relax

Junjun2016: fixed

PR #252, opened 2020-11-13
Support soft cross entropy loss

Junjun2016:
Support soft cross entropy loss


yiminglin-ai:
How is the performance compared with standard (hard) CrossEntropy?

PR #268, opened 2020-11-19
[New model] Support MobileNetV3

yamengxi:


yamengxi:
I built the network the same as https://github.com/ekzhang/fastseg, but get a lower result as following:
| config | mIoU(my result) | mIoU(fastseg) | mIoU(original paper) |
|-----------|----------|----------|----------|
| configs/mobilenet_v3/lraspp_m-v3-d8_512x1024_160k_cityscapes.py | 67.56 | 72.3 | 72.36 |
| configs/mobilenet_v3/lraspp_m-v3s-d8_512x1024_160k_cityscapes.py | 61.43 | 67.4 | 68.38 |

I don't know why my training results are lower than original paper (about 5~6 points).
Can anyone give me some advice and help?

hellock:
Task linked: [CU-fjq4w0 MobileNetV3](https://app.clickup.com/t/fjq4w0)

.gitignore
```diff
@@ -112,6 +112,8 @@ data
 *.pkl.json
 *.log.json
 work_dirs/
+old_work_dirs/
+configs/mobilenet_v3_test/
```
xvjiarui, line 116: These modification is not necessary.

mmseg/models/decode_heads/lraspp_head.py
```diff
@@ -0,0 +1,90 @@
...
+                1,
+                act_cfg=dict(type='Sigmoid'),
+                bias=False))
+
+    def forward(self, inputs):
+        """Forward function."""
+        inputs = self._transform_inputs(inputs)
+
+        x = inputs[-1]
+        assert x.size(2) >= 49 and x.size(3) >= 49
```
xvjiarui, line 73: Why is 49?

.gitignore
yamengxi: OK, I will delete it in the final version.

mmseg/models/decode_heads/lraspp_head.py
yamengxi: It is proposed in the paper.
image

yamengxi: We have to make sure that the input size is not less than 49.

xvjiarui: Will PyTorch raise the error if we don't assert? Or we may add comments here.

xvjiarui (approved):

PR #334, opened 2021-01-07
[Bug Fix] Fix TTA resize scale

yamengxi:


mmseg/datasets/pipelines/transforms.py
```diff
@@ -157,7 +157,7 @@ def _random_scale(self, results):
         if self.ratio_range is not None:
             if self.img_scale is None:
                 scale, scale_idx = self.random_sample_ratio(
-                    results['img'].shape[:2], self.ratio_range)
+                    results['img'].shape[:2:-1], self.ratio_range)
```
xvjiarui, line 160:
```suggestion
                    h, w = results['img'].shape[:2]
                    (w, h), self.ratio_range)
```
is more intuitive.

xvjiarui (approved):

PR #336, opened 2021-01-08
remove the model with 0 ckpt and exclude the papers without proposing any model

congee524:


congee524:
to modify model zoo statistics part of mmsegmentation doc

xvjiarui (approved):

PR #374, opened 2021-02-08
Fixed Bug in Demo/MMSegmentation_Tutorial.ipynb 

know-one-1:
In build_segmentor line parameters "train_cfg=cfg.train_cfg, test_cfg=cfg.test_cfg" was creating error, because it has been moved to model. Deleting these parameters from function solves the error.


xvjiarui:
Fixed! Thx!

PR #392, opened 2021-03-01
[new feature] Add albumentation

24hours:


xvjiarui:
Hi @24hours 
Thanks for the contribution. And sorry for the late reply. Since Albu is not a requirement of our repo, we may make it none by default. We may have something like [this](https://github.com/open-mmlab/mmdetection/blob/dbc6b67446d3898461d0a5d5cef7d112e710db0a/mmdet/datasets/pipelines/transforms.py#L17-L22).

24hours:
Hi, 

> Hi @24hours
> Thanks for the contribution. And sorry for the late reply. Since Albu is not a requirement of our repo, we may make it none by default. We may have something like [this](https://github.com/open-mmlab/mmdetection/blob/dbc6b67446d3898461d0a5d5cef7d112e710db0a/mmdet/datasets/pipelines/transforms.py#L17-L22).

I have made the relevant change, please check.

maisonhai3:
> Hi @24hours
> Thanks for the contribution. And sorry for the late reply. Since Albu is not a requirement of our repo, we may make it none by default. We may have something like [this](https://github.com/open-mmlab/mmdetection/blob/dbc6b67446d3898461d0a5d5cef7d112e710db0a/mmdet/datasets/pipelines/transforms.py#L17-L22).

I think it better we have albumentation already implanted in mmsegmentation,
make it ready-to-use,
than making users manually copy-pasted pieces code from mmdetection.

Junjun2016:
Hi @24hours 
Thanks for your contribution.
Please fix the conflicts.

xiexinch:
https://github.com/open-mmlab/mmsegmentation/pull/2710

PR #400, opened 2021-03-04
Support fcn dilate 6

xiexinch:


xvjiarui:
We may rename the file head to `fcn-d6_r50-d16...`. We could do cityscapes only for this PR.

xvjiarui:
We may also add config for resnet 101.

mmseg/models/decode_heads/fcn_head.py
```diff
@@ -17,28 +17,32 @@ class FCNHead(BaseDecodeHead):
         kernel_size (int): The kernel size for convs in the head. Default: 3.
         concat_input (bool): Whether concat the input and output of convs
             before classification layer.
+        dilation (int): Spacing between kernel elements. Default: 1.
```
xvjiarui, line 20:
```suggestion
        dilation (int): The dilation rate for convs in the head. Default: 1.
```

xvjiarui (approved):

PR #405, opened 2021-03-10
[feature]: add-focal-loss

yinchimaoliang:
Add focal_loss in losses module.


xvjiarui:
Hi @yinchimaoliang 
Thanks for the contribution. 
We provide faster focal loss [here](https://github.com/open-mmlab/mmcv/blob/master/mmcv/ops/focal_loss.py). We may also support it.

daleydeng:
> Hi @yinchimaoliang
> Thanks for the contribution.
> We provide faster focal loss [here](https://github.com/open-mmlab/mmcv/blob/master/mmcv/ops/focal_loss.py). We may also support it.

But how to handle ignore_label=255 in it? there seems no such feature.
the proposed version by @yinchimaoliang has such thing.

xiximelon:
Thanks a lot ! I can use it to run my code

PR #426, opened 2021-03-16
add plot_logs tool

xiexinch:
Support plot logs by matplotlib.


xvjiarui:
We may also update the doc.

xvjiarui (approved):

PR #436, opened 2021-03-22
[Feature] Add support for Batch Inference

daavoo:


sennnnn:
We plan to restart this pr after merge of #709

PR #438, opened 2021-03-24
Use MMCV's EvalHook in MMSegmentation

xiexinch:
Apply EvalHook and DistEvalHook implemented in MMCV open-mmlab/mmcv#739 .


xvjiarui:
unit test failed.

xvjiarui:
We may directly inherit from mmcv instead of define all of them.

xvjiarui (approved):

PR #417, opened 2021-03-13
[Bug fixed]Fix dice_loss errors

xiexinch:
Fix training errors  #411

PR #440, opened 2021-03-25
refactor test organization

xvjiarui:


tests/test_models/test_losses/test_dice_loss.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+import torch
+
+
+def test_dice_lose():
```
xiexinch, line 5: I have modified the test function at #417 ,  could you update it ?

PR #474, opened 2021-04-13
Quik script for pascal context dataset conversion

sennnnn:
* Using trainval_merged.json to get train_ids and val_ids;

* Extracting segmentation map from raw mat file;

* Improve the conversion time from half an hour to one minute;

### DIFF
--- a/tools/convert_datasets/quik_pascal_context.py
+++ b/tools/convert_datasets/quik_pascal_context.py
@@ -1,0 +1,125 @@
+import argparse
+import json
+import os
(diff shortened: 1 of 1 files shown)


sennnnn:
> * Using trainval_merged.json to get train_ids and val_ids;
> * Extracting segmentation map from raw mat file;
> * Improve the conversion time from half an hour to one minute;

The masks generated by quik convert script are different from masks directly extracted from trainval_merged.json.

PR #512, opened 2021-04-23
add upsample neck

xiexinch:
A neck structure connect vit backbone and decoder_heads.
Default output of vit backbone is a downsample 16x feature map, we design a structure that convert the feature map to [32x, 16x, 8x, 4x] feature map list.


mmseg/models/necks/upsample_neck.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+import torch.nn as nn
+import torch.nn.functional as F
+from mmcv.cnn import ConvModule
+
+from ..builder import NECKS
+
+
+@NECKS.register_module()
+class UpsampleNeck(nn.Module):
```
xvjiarui, line 10: We may rename it to something like `MultiLevelNeck` since we would also perform downsampling.

xvjiarui (approved):

mmseg/models/necks/multilevel_neck.py
```diff
@@ -0,0 +1,70 @@
...
+
+@NECKS.register_module()
+class MultiLevelNeck(nn.Module):
+    """MultiLevelNeck.
+
+    A neck structure connect vit backbone and decoder_heads.
+    Args:
+        in_channels (List[int]): Number of input channels per scale.
+        out_channels (int): Number of output channels (used at each scale).
+        scales (List[int]): Scale factors for each input feature map.
```
xvjiarui, line 16: Default is missing.

xvjiarui: We change fix this in the new PR.

PR #513, opened 2021-04-25
[Enhance] Support reading class_weight from file in loss functions to help MMDet3D

Wuziyi616:
Hi, I am Ziyi, a developer from MMDet3D. Currently we are implementing 3D segmentation models, which require using loss functions from MMSeg (e.g. CrossEntropy). It's very common to use class_weight in 3D Seg, and in some scenarios we need to load class_weight from files. So I modify the code and support this feature. Hope you can approve it. Thanks.

### DIFF
--- a/mmseg/models/losses/cross_entropy_loss.py
+++ b/mmseg/models/losses/cross_entropy_loss.py
@@ -146,8 +146,8 @@ class CrossEntropyLoss(nn.Module):
             Defaults to False.
         reduction (str, optional): . Defaults to 'mean'.
             Options are "none", "mean" and "sum".
-        class_weight (list[float], optional): Weight of each class.
-            Defaults to None.
+        class_weight (list[float] | str, optional): Weight of each class. If in
+            str format, read them from a file. Defaults to None.
         loss_weight (float, optional): Weight of the loss. Defaults to 1.0.
     """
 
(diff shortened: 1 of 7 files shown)


mmseg/models/losses/cross_entropy_loss.py
```diff
@@ -172,6 +172,28 @@ def __init__(self,
...
+        import mmcv
+
+        if isinstance(class_weight, str):
+            # take it as a file path
+            if class_weight.endswith('.npy'):
+                class_weight = np.load(class_weight)
+            else:
+                file_format = class_weight.split('.')[-1]
+                assert file_format in ['json', 'yaml', 'pkl'], \
+                    f'unsupported class_weight file type {file_format}'
```
xvjiarui, line 192: This part is already handled by MMCV [here](https://github.com/open-mmlab/mmcv/blob/master/mmcv/fileio/io.py#L37)

mmseg/models/losses/cross_entropy_loss.py
```diff
@@ -172,6 +172,28 @@ def __init__(self,
         else:
             self.cls_criterion = cross_entropy
 
+    def _get_class_weight(self, class_weight):
```
xvjiarui, line 175: Could we make this an utility function that could be shared by all losses?

mmseg/models/losses/cross_entropy_loss.py
Wuziyi616: Great to hear.

mmseg/models/losses/cross_entropy_loss.py
Wuziyi616: Yes, good idea!

Wuziyi616:
This PR is ready for review @xvjiarui. Thanks!

xvjiarui (approved):

PR #514, opened 2021-04-25
[Enhance] Replace data_dict calling 'img' key to support MMDet3D

Wuziyi616:
Hi, I am Ziyi, a developer from MMDet3D. In order to train/test 3D seg algorithms, I need to call functions from MMSeg. However, some functions in MMSeg calls `data['img']`, which causes `KeyError` because we don't have `img`. I replace these calling with some workarounds and make it compatibility with MMDet3D. Hope you can approve it. Thanks.

### DIFF
--- a/mmseg/apis/test.py
+++ b/mmseg/apis/test.py
@@ -92,12 +92,13 @@ def single_gpu_test(model,
             if efficient_test:
                 result = [np2tmp(_) for _ in result]
             results.extend(result)
+            batch_size = len(result)
         else:
             if efficient_test:
                 result = np2tmp(result)
             results.append(result)
+            batch_size = 1
 
-        batch_size = data['img'][0].size(0)
         for _ in range(batch_size):
             prog_bar.update()
     return results
--- a/mmseg/apis/test.py
+++ b/mmseg/apis/test.py
@@ -97,7 +97,7 @@ def single_gpu_test(model,
                 result = np2tmp(result)
             results.append(result)
 
-        batch_size = len(data['img_metas'][0])
+        batch_size = len(result)
         for _ in range(batch_size):
             prog_bar.update()
     return results
--- a/mmseg/apis/test.py
+++ b/mmseg/apis/test.py
@@ -92,13 +92,12 @@ def single_gpu_test(model,
             if efficient_test:
                 result = [np2tmp(_) for _ in result]
             results.extend(result)
-            batch_size = len(result)
         else:
             if efficient_test:
                 result = np2tmp(result)
             results.append(result)
-            batch_size = 1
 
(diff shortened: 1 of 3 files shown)


mmseg/apis/test.py
```diff
@@ -92,12 +92,13 @@ def single_gpu_test(model,
             if efficient_test:
                 result = [np2tmp(_) for _ in result]
             results.extend(result)
+            batch_size = len(result)
         else:
             if efficient_test:
                 result = np2tmp(result)
             results.append(result)
+            batch_size = 1
```
xvjiarui, line 100: The batch size may not be 1 under all cases.

mmseg/apis/test.py
```diff
@@ -92,12 +92,13 @@ def single_gpu_test(model,
...
 
-        batch_size = data['img'][0].size(0)
```
xvjiarui, line 100: Will `data['img_meta']` do the trick?

Wuziyi616: Thanks for your reply. I am actually not quite sure about the data's shape here. Does `data['img_meta']` looks like `List[List[dict]]`, where the outer is different augmentation and inner is a batch? So I should use `len(data['img_metas'][0])` here?

xvjiarui: Yes. It is the same as MMDet.

Wuziyi616: Fix with `data['img_metas']`.

tests/test_eval_hook.py
```diff
@@ -16,7 +16,7 @@
 class ExampleDataset(Dataset):
 
     def __getitem__(self, idx):
-        results = dict(img=torch.tensor([1]), img_metas=dict())
+        results = dict(img=torch.tensor([1]), img_metas=[dict()])
```
Wuziyi616, line 19: To pass the unit test, we need `img_metas` to be wrapped in a list.

Wuziyi616:
This PR is ready for review @xvjiarui. Thanks!

mmseg/apis/test.py
```diff
@@ -97,7 +97,7 @@ def single_gpu_test(model,
                 result = np2tmp(result)
             results.append(result)
 
-        batch_size = data['img'][0].size(0)
+        batch_size = len(result)
```
Wuziyi616, line 100: mmcv issue [here](https://github.com/open-mmlab/mmcv/pull/987)

Wuziyi616: Refer to open-mmlab/mmcv#987

Wuziyi616, line 100: Refer to open-mmlab/mmcv#987

xvjiarui (approved):

PR #530, opened 2021-05-06
Add option for output shape of ViT

sennnnn:
Add arg:final_reshape for vit to control if converting output feature information from NLC to NCHW.

### DIFF
--- a/mmseg/models/backbones/vit.py
+++ b/mmseg/models/backbones/vit.py
@@ -306,7 +306,8 @@ class VisionTransformer(nn.Module):
                 with_cp=with_cp) for i in range(depth)
         ])
 
-        assert out_shape in ['NLC', 'NCHW']
+        assert out_shape in ['NLC',
+                             'NCHW'], 'output shape must be "NLC" or "NCHW".'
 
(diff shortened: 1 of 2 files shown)


mmseg/models/backbones/vit.py
```diff
@@ -261,6 +263,7 @@ def __init__(self,
                  act_cfg=dict(type='GELU'),
                  norm_eval=False,
                  final_norm=False,
+                 final_reshape=True,
```
xvjiarui, line 266:
```suggestion
                 out_shape=`NCHW`,
```
We may rename it as `out_shape`.

mmseg/models/backbones/vit.py
```diff
@@ -305,6 +308,7 @@ def __init__(self,
 
         self.interpolate_mode = interpolate_mode
         self.final_norm = final_norm
+        self.final_reshape = final_reshape
```
xvjiarui, line 311: We may also add `out_shape in ['NCHW', 'NLC']`.

xvjiarui (approved):

PR #535, opened 2021-05-08
[Fix] Fix error when class_weight file doesn't exist

Wuziyi616:
In MMDet3D, there is a config file that builds a loss reading `class_weight` from a file. However, this file is generated via data pre-processing, thus doesn't exist on the GitHub repo. So the pytest will report an error in GitHub CI. This PR is a workaround for this error, which uses `None` if class_weight file doesn't exist and raise a warning.

### DIFF
--- a/mmseg/models/losses/utils.py
+++ b/mmseg/models/losses/utils.py
@@ -1,4 +1,5 @@
 import functools
+import warnings
 
 import mmcv
 import numpy as np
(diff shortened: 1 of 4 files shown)


Wuziyi616:
Linting error is caused by network connection failure. Can we re-run it?

xvjiarui:
If `class_weight` file doesn't exist, we'd better raise an error. The warning could be ignored in many cases. Do we have other work around for CI?

Wuziyi616:
Yes, I agree with you. Maybe I should hard-code the class_weight in the config file. Then do you think the `mmcv.check_file_exist(class_weight)` sentence is necessary? If not, I think you can just close this PR.

PR #547, opened 2021-05-12
fix verify bugs

grimoire:
Hi
This PR fixes `shape mismatch` when verifying the `whole mode` model in `pytorch2onnx`.

### DIFF
--- a/tools/pytorch2onnx.py
+++ b/tools/pytorch2onnx.py
@@ -96,12 +96,12 @@ def _prepare_input_img(img_path,
     return mm_inputs
 
 
-def _update_input_img(img_list, img_meta_list, update_origin=False):
+def _update_input_img(img_list, img_meta_list, update_ori_shape=False):
     # update img and its meta list
     N, C, H, W = img_list[0].shape
     img_meta = img_meta_list[0][0]
     img_shape = (H, W, C)
-    if update_origin:
+    if update_ori_shape:
         ori_shape = img_shape
     else:
         ori_shape = img_meta['ori_shape']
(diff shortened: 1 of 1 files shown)


tools/pytorch2onnx.py
```diff
@@ -96,13 +96,16 @@ def _prepare_input_img(img_path,
     return mm_inputs
 
 
-def _update_input_img(img_list, img_meta_list):
+def _update_input_img(img_list, img_meta_list, update_origin=False):
```
xvjiarui, line 99: What does `update_origin` means?

tools/pytorch2onnx.py
```diff
@@ -220,7 +223,7 @@ def pytorch2onnx(model,
 
             # update img_meta
             img_list, img_meta_list = _update_input_img(
-                img_list, img_meta_list)
+                img_list, img_meta_list, test_mode == 'whole')
```
xvjiarui, line 226: Why arg is `test_mode == 'whole'`?

grimoire: The input tensor has been interpolated with scale_factor=1.5 to test dynamic shape support.
We only support dynamic shape on these `whole mode` models.

tools/pytorch2onnx.py
grimoire: `ori_shape` in img_meta is used to rescale output to origin shape.
`_update_input_img` will update `img_meta_list` according to `img_list`. And this flag is used to control if the segmentator needs to rescale according to the input tensor shape.

xvjiarui (approved):

PR #549, opened 2021-05-12
[Feature] support mim

xiexinch:
Reference

1. Add metafile, ref https://github.com/open-mmlab/mmdetection3d/blob/master/model_zoo.yml

2. Add MANIFEST.in,  ref https://github.com/open-mmlab/mmdetection3d/blob/master/MANIFEST.in

3. Set `include_package_data=True` at setup.py, ref https://github.com/open-mmlab/mmdetection3d/blob/master/setup.py#L151

4. Add mminstall.txt to `requirements`, ref https://github.com/open-mmlab/mmdetection3d/blob/master/requirements/mminstall.txt and https://github.com/open-mmlab/mmclassification/blob/master/requirements/mminstall.txt


xvjiarui:
Example for Swin Transformer https://github.com/open-mmlab/mim-example/pull/1

configs/ccnet/metafile.yml
```diff
@@ -0,0 +1,231 @@
...
+          mIoU: 42.08
+    Weights: https://download.openmmlab.com/mmsegmentation/v0.5/ccnet/ccnet_r50-d8_512x512_160k_ade20k/ccnet_r50-d8_512x512_160k_ade20k_20200616_084435-7c97193b.pth
+    Config: configs/ccnet/ccnet_r50-d8_512x512_160k_ade20k.py
+
+
+
+  - Name: ccnet_r101-d8_512x512_160k_ade20k
+    In Collection: CCNet
+    Metadata:
+      inference time (fps):
```
xvjiarui, line 168: Some FPS is missing.

xvjiarui (approved):

PR #531, opened 2021-05-06
[Feature] Official implementation of SETR

sixiaozheng:
This is from the **official implementation** of SETR

PR #520, opened 2021-04-26
add configs for vit backbone plus decode_heads

xiexinch:

PR #571, opened 2021-05-25
[Refactor] Using mmcv transformer bricks to refactor vit.

sennnnn:
The foundation of this PR:
mmcv: https://github.com/open-mmlab/mmcv/pull/978 (merged)
mmclassification: https://github.com/open-mmlab/mmclassification/pull/295


xvjiarui:
We may exclude SETR from this PR. We will refactor #531 and #520

mmseg/models/backbones/vit.py
```diff
@@ -344,27 +333,8 @@ def init_weights(self, pretrained=None):
...
-                    kaiming_init(m.weight, mode='fan_in')
-                    if m.bias is not None:
-                        constant_init(m.bias, 0)
-                elif isinstance(m, (_BatchNorm, nn.GroupNorm, nn.LayerNorm)):
-                    constant_init(m.bias, 0)
-                    constant_init(m.weight, 1.0)
-        else:
-            raise TypeError('pretrained must be a str or None')
+            # Modified from ClassyVision
+            nn.init.normal_(self.pos_embed, std=0.02)
```
xvjiarui, line 337: Why remove the initialization?

mmseg/models/backbones/vit.py
```diff
@@ -330,10 +299,17 @@ def init_weights(self, pretrained=None):
             else:
                 state_dict = checkpoint
 
+            if 'rwightman/pytorch-image-models' in pretrained:
```
xvjiarui, line 302: If user loaded the weight from timm and would like to init the model with path, the condition does not hold.

.github/workflows/build.yml
```diff
@@ -37,19 +37,26 @@ jobs:
         include:
           - torch: 1.3.0+cpu
             torchvision: 0.4.1+cpu
+            torch_version: 1.3.0
```
xvjiarui, line 40: Reset this file.

mmseg/models/backbones/vit.py
```diff
@@ -1,294 +1,257 @@
...
-            Default: None
+        num_fcs (int): The number of fully-connected layers for FFNs.
+            Default: 2.
+        norm_eval (bool): Whether to set norm layers to eval mode, namely,
+            freeze running stats (mean and var). Note: Effect on Batch Norm
+            and its variants only. Default: False.
+        with_cp (bool): Use checkpoint or not. Using checkpoint will save
+            some memory while slowing down the training speed. Default: False.
+        pretrain_style (str): Choose to use timm or mmcls pretrain weights.
+            Default: timm.
```
xvjiarui, line 200: We should explain what options are supported, and add assert.

xvjiarui (approved):

PR #575, opened 2021-05-30
[feature]: Able to use save_best option

yinchimaoliang:
Change `eval_hooks.py`, able to use `save_best` while training.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -149,6 +149,8 @@ def main():
             PALETTE=datasets[0].PALETTE)
     # add an attribute for visualization convenience
     model.CLASSES = datasets[0].CLASSES
+    # passing checkpoint meta for saving best checkpoint
+    meta.update(cfg.checkpoint_config.meta)
     train_segmentor(
         model,
         datasets,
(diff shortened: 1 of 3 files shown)


tools/train.py
```diff
@@ -149,7 +149,6 @@ def main():
             PALETTE=datasets[0].PALETTE)
     # add an attribute for visualization convenience
     model.CLASSES = datasets[0].CLASSES
-    meta.update(cfg.checkpoint_config.meta)
```
yinchimaoliang: This seems to be necessary, without updating meta, the best ckpt can not be used to test.

xvjiarui: Why we need to update meta?

yinchimaoliang: Right now, meta missing some keys such as `CLASSES` and `PALETTE`, so the best ckpt it saves won't have these keys, which will cause failure during test time. Updating meta will solve the problem.

xvjiarui: I see. Reverted.

xvjiarui (approved):

PR #587, opened 2021-06-06
[Feature] Support coco-stuff dataset

DerrickWang005:
update dataset_prepare.md


mmseg/datasets/coco_stuff.py
```diff
@@ -0,0 +1,236 @@
...
+        [192, 192, 224],
+        [64, 32, 64],
+        [64, 192, 128],
+        [64, 192, 224],
+        [192, 160, 64],
+        [64, 192, 0],
+        [192, 192, 96],
+        [192, 96, 192],
+        [64, 64, 128],
+        [64, 64, 96],
```
xvjiarui, line 228:
```suggestion
        [64, 64, 96]
```

xvjiarui:
Hi @DerrickWang005 

Thanks for the nice PR. We will review and benchmark it soon.

DerrickWang005:
As for the palette, I adopt the PASCAL VOC style. Maybe you can use the official style.😂

xvjiarui:
> As for the palette, I adopt the PASCAL VOC style. Maybe you can use the official style.😂

It would be nice if we could use the official palette.

DerrickWang005:
I found COCO Stuff dataset has two versions, 10k and 164k. After updating that, I will create a new PR ASAP.

PR #609, opened 2021-06-18
[Fix] Fix some vit init bugs

sennnnn:
1. Fix some new style parameters init bug for vit.

2. Add some unit test for vit to improve code cov.

PR #610, opened 2021-06-19
[Feature] support NLC output shape

xiexinch:


sennnnn:
The changes of this PR are moved to #609.

PR #634, opened 2021-06-23
[Feature] tversky loss

MengzhangLI:
Adding tversky loss.

Not too much difference with original dice_loss.py, just adding weight on false positives and false negatives.


MengzhangLI:
Because creating new PR, so this old version is closed.

PR #635, opened 2021-06-23
[Fix] Add setr & vit msg.

sennnnn:
## Motivation

1. The lack of SETR & ViT in the model list. 

2. The SETRUPHead still use deprecated init_weights.

## Modification

1.  Add SETR & ViT url in the main readme doc.

2. Refactor init of SETRUPHead.

### DIFF
--- a/mmseg/models/decode_heads/setr_up_head.py
+++ b/mmseg/models/decode_heads/setr_up_head.py
@@ -35,7 +35,7 @@ class SETRUPHead(BaseDecodeHead):
         assert isinstance(self.in_channels, int)
 
         self.init_cfg = [
-            dict(dict='Constant', val=1.0, bias=0, layer='LayerNorm')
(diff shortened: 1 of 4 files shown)


mmseg/models/decode_heads/setr_up_head.py
```diff
@@ -25,6 +25,7 @@ def __init__(self,
                  num_convs=1,
                  up_scale=4,
                  kernel_size=3,
+                 init_cfg=None,
```
Junjun2016, line 28: We can set the default `init_cfg` to implement the same weights initialization as that in `init_weights`

Junjun2016: And we can set it to `None` in configs if SETR does not use it.

xvjiarui (approved):

Issue #548, opened 2021-05-12
weights of DMNet?


Anikily:
I can't download the weights of DMNet, could you help me?

Junjun2016:
Hi @Anikily 
We will fix it ASAP.

MengzhangLI:
Prepare to fix it.
Please check its related PR: https://github.com/open-mmlab/mmsegmentation/pull/660

Best,

MengzhangLI:
Hi, @Anikily 

This problem is fixed, we just checked several new links of DMNet model and their results is correct.

Please, check it out.
z

Issue #415, opened 2021-03-13
The UNet model link is error


sshuair:
the following link access is denied.
Backbone: UNet-S5-D16
Head: FCN
https://download.openmmlab.com/mmsegmentation/v0.5/unet/fcn_unet_s5-d16_64x64_40k_drive/fcn_unet_s5-d16_64x64_40k_drive_20201223_191051-26cee593.pth

xvjiarui:
Hi @sshuair 
Sorry about that. We will fix it soon.

MengzhangLI:
Still not working.

We would fix it ASAP.

MengzhangLI:
Prepare to fix it.
Please check its related PR: https://github.com/open-mmlab/mmsegmentation/pull/660

Best,

MengzhangLI:
Hi, @sshuair 

This problem is fixed, we just checked several new links of UNet (FCN) model and their results is correct.

Please, check it out.

PR #660, opened 2021-07-01
[Fix] Fix wrong links of .pth and .json file in DMNet and UNet(FCN) README.md

MengzhangLI:
Fix wrong links of .pth and .json file of DMNet and UNet(FCN) in `./configs/` folder.

Just as two related issues which reflect unsuccessful link of DMNet (https://github.com/open-mmlab/mmsegmentation/issues/548) and UNet drive FCN S5-D16 (https://github.com/open-mmlab/mmsegmentation/issues/415), we fix the correct link of those models.

Best,


Junjun2016:
Hi @MengzhangLI 
Could you please test several models

Junjun2016 (approved):

Issue #695, opened 2021-07-12
`init_cfg` type pretrained is not applied in some backbones


puhsu:
**Checklist**

1. I have searched related issues but cannot get the expected help. ✅ 
2. The bug has not been fixed in the latest version. ✅ 

**Describe the bug**

When using `init_cfg(type='Pretrained', checkpoint='...')` option for resnet backbone (and some others too), the `init_cfg` variable is not set for the backbone, as a result pretrained weight are not loaded to the backbone:

https://github.com/open-mmlab/mmsegmentation/blob/f5be6e2617aae15915060598b3f8004f4d9aeea1/mmseg/models/backbones/resnet.py#L412-L439

Am I right that there should be an else clause like this?

```python
if init_cfg is None:
    ...
else:
    self.init_cfg = init_cfg
```

If this is a bug, I would be happy to create a pr and fix it.

Junjun2016:
Hi @puhsu 
Thanks for your feedback!
Now, we pass backbone pretrained weights through [`pretrained`](https://github.com/open-mmlab/mmsegmentation/blob/master/configs/_base_/models/fcn_r50-d8.py#L5) argument to compatible with the old version, which will be deprecated in the future.
The `init_cfg` is suggested!

puhsu:
Yes, it works with `pretrained` now. But I used my own config with `init_cfg` instead of `pretrained` (I saw the deprecation warning and decided to go with `init_cfg`), but the checkpoint is not loading this way (`init_cfg` is not set for modules). This should probably be fixed before removing the `pretrained` option

Junjun2016:
Could you please provide the whole config?
And we will test with it.
You can also check [`BaseModule`](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/base_module.py#L50) and [`PretrainedInit`](https://github.com/open-mmlab/mmcv/blob/master/mmcv/cnn/utils/weight_init.py#L386) in MMCV

sennnnn:
pretrained argument will be deprecated in the future. So, it may be better to use PretrainedInit to load pretrained. If you have some special ckpt load action, you can overload init_weights of backbone.

puhsu:
@Junjun2016 yep. Here is the model [config](https://gist.github.com/puhsu/4d8168556bb0d60f99eb6d91addb5bfd). I just swapped pretrained for init_cfg.

[...]

PR #697, opened 2021-07-12
Fix init_cfg in resnet backbone

puhsu:
Here is the quick fix for resnet backbone init_cfg option I talked about in #695


Junjun2016 (approved):
Thanks for your PR!

xvjiarui (approved):

PR #728, opened 2021-07-26
[Enhancement] Add a more reasonable model gather tool.

sennnnn:
## Why provide a new model gather tool?

1. The old model gather tool only support last ckpt extraction;

2. The old model gather tool don't support window platform;

3. The old model gather tool may be unfriendly to users;

## TODO

- [x] Support extract best or last ckpt and rename it to canonical format;

- [ ] Support generating a structured file which can be easier to fetch evaluation or training info;

...

### DIFF
--- a/.dev/gather_models.py
+++ b/.dev/gather_models.py
@@ -1,16 +1,19 @@
 import argparse
-import glob
+import hashlib
 import json
 import os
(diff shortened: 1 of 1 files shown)


sennnnn:
We will move it to a general pr to add develop tool.

PR #734, opened 2021-07-27
[Fix] Use LooseVersion to compare different version instead of string

Junjun2016:
## Motivation

Using string to compare versions is not safe.
'1.12.0' >= '1.8.0' will reture `False`

## Modification

`torch.__version__ >= '1.8.0'` --> `LooseVersion(torch.__version__) >= LooseVersion('1.8.0')`

## BC-breaking (Optional)

No


mmseg/datasets/builder.py
```diff
@@ -133,7 +134,7 @@ def build_dataloader(dataset,
         worker_init_fn, num_workers=num_workers, rank=rank,
         seed=seed) if seed is not None else None
 
-    if torch.__version__ >= '1.8.0':
+    if LooseVersion(torch.__version__) >= LooseVersion('1.8.0'):
```
zhouzaida, line 137: please use `digit_version` provided in mmcv to check version

PR #750, opened 2021-08-04
[Enhancement] support user inherit cityscapes dataset

xiexinch:

## Motivation

Support user custom dataset by inheriting `Cityscapes` class, the original implementation requires user's image suffix same with cityscapes, it's not flexible.

## Modification

Set default value to the constructor function parameter.

## Use cases (Optional)

```python
@DATASETS.register_module()
class MyDataset(CityscapesDataset):
    """MyDatasett."""

    def __init__(self, **kwargs):
        super().__init__(
            img_suffix='_train.png',
            seg_map_suffix='_gt_mask.png',
            **kwargs)
```

### DIFF
--- a/mmseg/datasets/cityscapes.py
+++ b/mmseg/datasets/cityscapes.py
@@ -30,10 +30,9 @@ class CityscapesDataset(CustomDataset):
                [0, 80, 100], [0, 0, 230], [119, 11, 32]]
 
     def __init__(self, **kwargs):
-        super(CityscapesDataset, self).__init__(
-            img_suffix='_leftImg8bit.png',
-            seg_map_suffix='_gtFine_labelTrainIds.png',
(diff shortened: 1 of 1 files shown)


mmseg/datasets/cityscapes.py
```diff
@@ -30,10 +30,9 @@ class CityscapesDataset(CustomDataset):
                [0, 80, 100], [0, 0, 230], [119, 11, 32]]
 
     def __init__(self, **kwargs):
```
xvjiarui, line 32:
```suggestion
    def __init__(self, img_suffix='_leftImg8bit.png', seg_map_suffix='_gtFine_labelTrainIds.png', **kwargs):
```

Junjun2016 (approved):

PR #754, opened 2021-08-05
[Fix] Correct Fastscnn config path

MengzhangLI:
Motivation:
Fix fastscnn config link, it is wrong right now.

Modification
Change link from
`https://github.com/open-mmlab/mmsegmentation/blob/master/configs/fast_scnn.py`
to 
`https://github.com/open-mmlab/mmsegmentation/blob/master/configs/fastscnn/fast_scnn_lr0.12_8x4_160k_cityscapes.py`


Junjun2016:
Hi @MengzhangLI 
Please fix the lint error!

PR #762, opened 2021-08-06
a new dataset 2021 "SEED" 2nd Jiangsu Big Data Development and Application Competition (Hualu Cup)-Medical and Health Track

Waterman0524:
This PR adds the following functionalities
    1.Add a conversion program for medical segmentation dataset

### DIFF
--- a/tools/convert_datasets/yiliao.py
+++ b/tools/convert_datasets/yiliao.py
@@ -1,0 +1,118 @@
+import argparse
+import os
+import os.path as osp
+import tempfile
+import zipfile
+import numpy as np
+import cv2
+import mmcv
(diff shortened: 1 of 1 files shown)


MengzhangLI:
Hi, thanks for your nice PR and we would reply in detail ASAP.

By the way, do you have your result using MMSegmentation?

As far as I know, the size of hualu 2020 data is large thus GPU memory error often happened. 

And we would better transform hualu data preparation into some dataset format like DRIVE, please check [here](https://github.com/open-mmlab/mmsegmentation/blob/master/docs_zh-CN/dataset_prepare.md).

Best,

PR #765, opened 2021-08-09
[Fix] Fix train

uni19:
## Motivation
    1. avoid overwriting setting from config
    2. using spawn to fix thread-pool leak when using lmdb

## Modification
    1. check `by_epoch` in eval config before setting it to default value
    2. change multi processing start method from fork to spawn


Junjun2016:
Hi @uni19 
Thanks for your nice PR!
We will review it ASAP!

Issue #758, opened 2021-08-06
With newest mmcv==1.3.10, text log bug


jianlong-yuan:
![image](https://user-images.githubusercontent.com/18137551/128449260-79b8b1dc-e5f8-47c4-b118-bb17787e6a32.png)

Junjun2016:
Hi @jianlong-yuan 
Please provide more details.

jianlong-yuan:
> Hi @jianlong-yuan
> Please provide more details.

What detail do you want?

Junjun2016:
Hi @jianlong-yuan 
Sorry for the bug!
We will fix it ASAP!

Junjun2016:
https://github.com/open-mmlab/mmcv/pull/1252

PR #766, opened 2021-08-09
[Fix] Fix the log error 

Junjun2016:
## Motivation

If the priority of EvalHook is higher than IterTimerHook, it will cause `KeyError: 'data_time'` (#758).
Since the `time` key will add to the output of log_buffer after IterTimeHook, the TextLoggerHook will print the `time` and  `data_time` at the same time.

This PR is based on https://github.com/open-mmlab/mmcv/pull/1252.
## Modification

Set the priority of EvalHook to LOW.

## BC-breaking (Optional)

No.

### DIFF
--- a/mmseg/apis/train.py
+++ b/mmseg/apis/train.py
@@ -107,7 +107,8 @@ def train_segmentor(model,
         eval_cfg = cfg.get('evaluation', {})
         eval_cfg['by_epoch'] = cfg.runner['type'] != 'IterBasedRunner'
         eval_hook = DistEvalHook if distributed else EvalHook
-        runner.register_hook(eval_hook(val_dataloader, **eval_cfg))
+        runner.register_hook(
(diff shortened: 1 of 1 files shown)


xvjiarui:
Fixed #758

xvjiarui (approved):

mmseg/apis/train.py
```diff
@@ -107,7 +107,8 @@ def train_segmentor(model,
         eval_cfg = cfg.get('evaluation', {})
         eval_cfg['by_epoch'] = cfg.runner['type'] != 'IterBasedRunner'
         eval_hook = DistEvalHook if distributed else EvalHook
-        runner.register_hook(eval_hook(val_dataloader, **eval_cfg))
+        runner.register_hook(
```
xvjiarui, line 110: We may add the link for related PRs.

Junjun2016: Done.
Please have a check again!

PR #778, opened 2021-08-11
[Fix] Update digit_version

Junjun2016:
## Motivation

The previous `digit_version`  has some problems, refers to open-mmlab/mmcv#1185 and open-mmlab/mmcv#1235

## Modification

Copy the `digit_version` from mmcv and use it to parse version.

## BC-breaking (Optional)

No

## Use cases (Optional)

## Checklist


Junjun2016:
@zhouzaida

mmseg/datasets/builder.py
```diff
@@ -10,6 +10,8 @@
 from mmcv.utils import Registry, build_from_cfg
 from torch.utils.data import DataLoader, DistributedSampler
 
+from mmseg import digit_version
```
ZwwWayne, line 13: should use the API in mmcv rather than mmseg.

tests/test_digital_version.py
```diff
@@ -0,0 +1,20 @@
+from mmseg import digit_version
```
ZwwWayne, line 1: test_digital_version -> test_digit_version

xvjiarui (approved):

PR #784, opened 2021-08-12
[Fix] Fix random behavior of update_model_index in pre-commit hook

Junjun2016:
## Motivation

The pre-commit hook update_model_index is observed to exit with code 1 (indicating files modified) randomly without actually modifying the metafiles. We found that this is caused by file I/O conflict when this pre-commit hook is running in a multi-threading manner.

## Modification

- Add `require_serial=true` in `.pre-commit-config.yaml` to prevent multi-threading that causes file I/O conflict
- Modify `update_model_index` in `md2yml.py` to prevent redundant file I/O
- Set `sort_keys=True` and update model `.yml` files

Related PR https://github.com/open-mmlab/mmpose/pull/866

## BC-breaking (Optional)

No.

### DIFF
--- a/.dev/md2yml.py
+++ b/.dev/md2yml.py
@@ -25,17 +25,23 @@ def dump_yaml_and_check_difference(obj, filename):
     Returns:
         Bool: If the target YAML file is different from the original.
     """
-    original = None
+
+    str_dump = mmcv.dump(obj, None, file_format='yaml', sort_keys=True)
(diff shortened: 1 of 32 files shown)


Junjun2016:
@zhouzaida

Junjun2016:
Hi @zhouzaida 
Could you please check the metafiles?

zhouzaida (approved):
LGTM

PR #866, opened 2021-09-10
[Fix] Fix iter bug when resuming checkpoint in distributed train

FreyWang:
## Motivation
As show in 
https://github.com/open-mmlab/mmcv/blob/b4bfeb53c57f2c843cb5015f9c0a2d1689dba9c4/mmcv/runner/base_runner.py#L379 ,
`cfg.gpu_ids` is used to calculate current iteration when resuming checkpoint in `epoch_based_runner` and `based_runner`, but it is `range(1)` in distributed training in default, which will lead to a wrong iteration. Here fix it to `range(world_size)`

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -7,7 +7,7 @@ import time
 
 import mmcv
 import torch
-from mmcv.runner import init_dist
+from mmcv.runner import init_dist, get_dist_info
 from mmcv.utils import Config, DictAction, get_git_hash
 
 from mmseg import __version__
@@ -94,6 +94,9 @@ def main():
     else:
         distributed = True
         init_dist(args.launcher, **cfg.dist_params)
+        # gpu_ids is used to calculate iter when resuming checkpoint,
+        _, world_size = get_dist_info()
+        cfg.gpu_ids = range(world_size)
 
     # create work_dir
     mmcv.mkdir_or_exist(osp.abspath(cfg.work_dir))
--- a/tools/train.py
+++ b/tools/train.py
@@ -7,7 +7,7 @@ import time
 
 import mmcv
 import torch
-from mmcv.runner import init_dist, get_dist_info
+from mmcv.runner import get_dist_info, init_dist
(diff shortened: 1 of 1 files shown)


Junjun2016:
Hi @FreyWang 
Thanks for your contribution again.
It can resume the right number of iterations for `epoch_based_runner` which inherits `resume` from `base_runner`.
But for `iter_based_runner` which overwrites the `resume` function, maybe we need to adjust the max number of iterations.

FreyWang:
> Hi @FreyWang
> Thanks for your contribution again.
> It can resume the right number of iterations for `epoch_based_runner` which inherits `resume` from `base_runner`.
> But for `iter_based_runner` which overwrites the `resume` function, maybe we need to adjust the max number of iterations.

I met the problem when using `epoch_based_runner`, It seems `iter_based_runner` override the function and have no probelm, but `epoch_based_runner` did not

Junjun2016:
> iter_based_runner

Do you have any idea to adjust the number of iterations for `iter_based_runner`?
We will discuss it in the next week.

Junjun2016 (approved):

xvjiarui (approved):

xvjiarui:
Should we add notes for resume with different number of GPUs somewhere?

Btw, I don't think resume with different batch size is a reasonable operator.

FreyWang:
> > Hi @FreyWang
> > Thanks for your contribution again.
> > It can resume the right number of iterations for epoch_based_runner which inherits resume from base_runner.
> > But for iter_based_runner which overwrites the resume function, maybe we need to adjust the max number of iterations.
> 
> Do you have any idea to adjust the number of iterations for `iter_based_runner`?
> We will discuss it in the next week.

Actually I agree that one resumed experiment should have same batch size with checkpoint.

FreyWang:
> Should we add notes for resume with different number of GPUs somewhere?
> 
> Btw, I don't think resume with different batch size is a reasonable operator.

I agree too, maybe batch size should be limited to be same

PR #882, opened 2021-09-17
[Enhancement] Using docker to skip CUDA installation in CI

Junjun2016:
Using docker to skip CUDA installation in CI.


.github/workflows/build.yml
```diff
@@ -62,98 +75,146 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install system dependencies
+        run: |
+          apt-get update && apt-get install -y ffmpeg libsm6 libxext6 git ninja-build libglib2.0-0 libsm6 libxrender-dev libxext6 python${{matrix.python-version}}-dev
+          apt-get clean
+          rm -rf /var/lib/apt/lists/*
```
xvjiarui, line 126: Are these necessary with container?

.github/workflows/build.yml
```diff
@@ -62,98 +75,146 @@ jobs:
...
       matrix:
         python-version: [3.7]
-        torch: [1.5.0+cu101, 1.7.0+cu101, 1.8.0+cu101, 1.9.0+cu102]
+        torch:
+          [
+            1.3.1,
+            1.5.1+cu101,
+            1.6.0+cu101,
+            1.7.0+cu101,
+            1.8.0+cu101,
```
xvjiarui, line 92:
```suggestion
            1.8.0+cu101
```

.github/workflows/build.yml
Junjun2016: Yes.

xvjiarui (approved):

PR #906, opened 2021-09-25
[Fix] Fix loss parse in val_step

acdart:
## Motivation

Add loss parse in val_step, so that logger can get loss value of validation dataset
 
## Modification

add loss parse and format output dict in val_step function

### DIFF
--- a/tests/test_models/test_segmentors/utils.py
+++ b/tests/test_models/test_segmentors/utils.py
@@ -101,13 +101,25 @@ def _segmentor_forward_train_test(segmentor):
         imgs, img_metas, gt_semantic_seg=gt_semantic_seg, return_loss=True)
     assert isinstance(losses, dict)
 
+    # Test train_step
+    data_batch = dict(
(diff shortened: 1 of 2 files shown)


Junjun2016:
Hi @acdart 
Thanks for your contribution.
It would be better to add some unittests if your time is available.

tests/test_models/test_segmentors/utils.py
```diff
@@ -101,6 +101,14 @@ def _segmentor_forward_train_test(segmentor):
         imgs, img_metas, gt_semantic_seg=gt_semantic_seg, return_loss=True)
     assert isinstance(losses, dict)
 
+    # Test val_step
+    with torch.no_grad():
+        segmentor.eval()
+        data_batch = dict(
+            img=imgs, img_metas=img_metas, gt_semantic_seg=gt_semantic_seg)
+        outputs = segmentor.val_step(data_batch)
+        assert isinstance(outputs, dict)
```
Junjun2016, line 110: Can also add asserts for the `keys` in `outputs`.

Junjun2016 (approved):
LGTM

PR #998, opened 2021-10-26
[Feature] Support TIMMBackbone

Junjun2016:
## Motivation

TIMM (https://github.com/rwightman/pytorch-image-models) contains a lot of backbones, which are not available in this repository. But mmsegmentation code is more flexible for training and adding new components for semantic segmentation. It would be convenient to have a wrapper here to use TIMM backbones.

Related PR：https://github.com/open-mmlab/mmclassification/pull/427

## Modification

This PR adds `TIMMBackbone` wrapper to use timm backbones.

## BC-breaking (Optional)

No

## Use cases (Optional)

To use `resnet18` from timm with pretrained weights and output stride 8, you need to use the following backbone config:
```python
backbone=dict(
    type="TIMMBackbone",
    model_name="resnet18",
    features_only=True,
    pretrained=True,   
    output_stride=8)
```
## Checklist


xvjiarui (approved):
#1004 is merge. Maybe we could try if TIMM works with 1.5.0

Junjun2016:
image

.github/workflows/build.yml
```diff
@@ -71,9 +71,17 @@ jobs:
         run: rm -rf .eggs && pip install -e .
       - name: Run unittests and generate coverage report
         run: |
+          pip install timm
           coverage run --branch --source mmseg -m pytest tests/
           coverage xml
           coverage report -m
+        if: ${{matrix.torch >= '1.5.0'}}
```
xvjiarui, line 78: Maybe we could undo the skipping

.github/workflows/build.yml
```diff
@@ -71,9 +71,17 @@ jobs:
...
+      - name: Skip timm unittests and generate coverage report
+        run: |
+          coverage run --branch --source mmseg -m pytest tests/ --ignore tests/test_models/test_backbones/test_timm_backbone.py
```
xvjiarui, line 81: Maybe we could undo the ignore

Junjun2016: TIMM does not support pt1.3

Junjun2016: `if: ${{matrix.torch < '1.5.0'}}` will skip the timm unittests.

PR #1039, opened 2021-11-14
[Enchance] Set a random seed when the user does not set a seed

Junjun2016:
## Motivation

Currently, when the seed is not set in the command line, the training process will not do anything with seed. This may cause a potential error in some cases (when different threads have different seeds).

- When the seed is not set in the args, it will randomly generate a seed and broadcast to all ranks.
- cfg.seed=seed

Related PR https://github.com/open-mmlab/mmdetection/pull/6354, https://github.com/open-mmlab/mmdetection/pull/6457, https://github.com/open-mmlab/mmdetection/issues/6339


RockeyCoss:
Please merge the master branch into your branch, thank you.

xvjiarui (approved):

PR #1119, opened 2021-12-09
[Fix] Unified links of readthedocs in docs

MengzhangLI:
Related PR: https://github.com/open-mmlab/mim/pull/89
https://github.com/open-mmlab/mmcv/blob/master/docs/conf.py#L101.

Note: 
(1) MMSegmentation has adopted `pytorch_sphinx_theme` from https://github.com/open-mmlab/mmsegmentation/pull/879.

(2) Currently, some repos do not have Chinese readthe docs, they are `mmhuman3d`, `mmflow` and `mmgeneration`, thus just keep their url. We may change it in the future.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -102,7 +102,7 @@ html_theme_options = {
             'children': [
                 {
                     'name': 'MMCV',
-                    'url': 'https://github.com/open-mmlab/mmcv',
+                    'url': 'https://mmcv.readthedocs.io/en/latest/',
                     'description': 'Foundational library for computer vision'
                 },
             ]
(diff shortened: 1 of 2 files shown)


Junjun2016 (approved):

Junjun2016 (changes requested):
The menu is not correct.
Please check the pre-builded readthedoc.

gaotongxiao:
These changes are no longer required and can be simplified as in https://github.com/open-mmlab/mmocr/pull/655

Junjun2016 (approved):

PR #1149, opened 2021-12-18
[Fix] Support MaskFormer.

wanghao9610:
## Motivation

Add MaskFormer to mmseg.

## Modification

* add 'ToMask' formatting operator to train pipeline to generate the 'gt_masks' and 'gt_labels'.
* change 'DefaultFormatBundle' class to adapt to new keys ' gt_masks' and 'gt_labels'.
* change 'EncoderDecoder' forward_train input elements (adding **kwargs to match the maskformer input demand).
* add 'maskformer_head', reference with the implementation of mmdet.
* change 'dice_loss' and 'focal_loss' reference with them on mmdet, the original dice_loss will lead the out_of_memory error because of the F.one_hot function when num_class is large.
* add 'plugin' subdirectory simply adding other modules, such as PixelDecoder, TransformerEncodePixelDecoder.
* add 'MaskHungarianAssigner' to do mask bipartite matching, reference with mmdet.
* add 'positional_encoding' and 'transformer', reference with mmdet.
* add some configs of maskformer:
	* maskformer_r50_512x512_80k_ade20k.py
	* maskformer_r50_512x512_160k_ade20k.py
	* maskformer_swin-t-p4-w7_512x512_160k_ade20k.py


MengzhangLI:
Based on our inner discussion, we would like to use MaskFormer module directly from MMDetection, rather than making a lot of code copy to MMSegmentation. We would support MaskFormer as soon as possible.

PR #1152, opened 2021-12-19
[Fix] Set random seed for `palette` if not given.

jbwang1997:

## Motivation

The palette may be different in each iteration if not specified.

## Modification

Add np.random.seed(42) in [base.py](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/models/segmentors/base.py#L212).

Following the method in mmdet.core.visualization.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmseg/models/segmentors/base.py
+++ b/mmseg/models/segmentors/base.py
@@ -245,9 +245,13 @@ class BaseSegmentor(BaseModule, metaclass=ABCMeta):
         seg = result[0]
         if palette is None:
             if self.PALETTE is None:
-                # Fix colors by setting random seed.
+                # Get random state before set seed,
+                # and restore random state later.
+                # Prevent loss of randomness.
+                # See: https://github.com/open-mmlab/mmdetection/issues/5844
                 state = np.random.get_state()
                 np.random.seed(42)
+                # random palette
                 palette = np.random.randint(
                     0, 255, size=(len(self.CLASSES), 3))
                 np.random.set_state(state)
--- a/mmseg/models/segmentors/base.py
+++ b/mmseg/models/segmentors/base.py
@@ -247,7 +247,8 @@ class BaseSegmentor(BaseModule, metaclass=ABCMeta):
             if self.PALETTE is None:
                 # Get random state before set seed,
                 # and restore random state later.
-                # Prevent loss of randomness.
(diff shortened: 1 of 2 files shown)


mmseg/models/segmentors/base.py
```diff
@@ -245,8 +245,12 @@ def show_result(self,
         seg = result[0]
         if palette is None:
             if self.PALETTE is None:
+                # Fix colors by setting random seed.
+                state = np.random.get_state()
```
Junjun2016, line 249: Can add more comments, refer to https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/visualization/image.py#L94-L96.

Junjun2016:
Hi @jbwang1997 
Thanks for your contribution.

mmseg/models/segmentors/base.py
MeowZheng: Also, add this line in comments “The palette may be different in each iteration if not specified.”

jbwang1997: > Also, add this line in comments “The palette may be different in each iteration if not specified.”

But if set np.random.seed(42), the random palette will be fixed in each iteration.

MeowZheng: Of course. My point is an additional straightforward explanation in comments for this modification.

AronLin (commented):
https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/datasets/custom.py#L352

Here is another random function, you can modify it at the same time.

mmseg/models/segmentors/base.py
```diff
@@ -245,8 +245,16 @@ def show_result(self,
         seg = result[0]
         if palette is None:
             if self.PALETTE is None:
+                # Get random state before set seed,
+                # and restore random state later.
+                # Prevent loss of randomness.
```
MeowZheng, line 250:
```suggestion
                # It will prevent loss of randomness, as the palette may be different in each iteration if not specified.
```

jbwang1997: OK, I will check it quickly.

MeowZheng (approved):

Junjun2016 (approved):
LGTM

PR #1161, opened 2021-12-24
fix dnl_head export onnx inference difference type Cast error

sshuair:
## Motivation

fix export onnx inference difference type Cast error.

## Modification

When use this class in model, if export to onnx, and use onnx runtime inference, It will case the Cast difference format error.

image

and the orginal onnx graph is:
- first Cast output is Float, but the `Pow` need Float64.
-  second Cast output is Int64, but the `Div` Op is need Float64

image

After modify the code used in this PR, the output onnx graph is 
image

## Checklist

### DIFF
--- a/mmseg/models/decode_heads/dnl_head.py
+++ b/mmseg/models/decode_heads/dnl_head.py
@@ -26,7 +26,9 @@ class DisentangledNonLocal2d(NonLocal2d):
         pairwise_weight = torch.matmul(theta_x, phi_x)
         if self.use_scale:
             # theta_x.shape[-1] is `self.inter_channels`
-            pairwise_weight /= torch.tensor(theta_x.shape[-1], dtype=torch.float)**torch.tensor(0.5)
+            pairwise_weight /= torch.tensor(
+                theta_x.shape[-1], dtype=torch.float
+            ) ** torch.tensor(0.5)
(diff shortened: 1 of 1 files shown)


mmseg/models/decode_heads/dnl_head.py
```diff
@@ -26,8 +26,9 @@ def embedded_gaussian(self, theta_x, phi_x):
         pairwise_weight = torch.matmul(theta_x, phi_x)
         if self.use_scale:
             # theta_x.shape[-1] is `self.inter_channels`
-            pairwise_weight /= theta_x.shape[-1]**0.5
-        pairwise_weight /= self.temperature
+            pairwise_weight /= torch.tensor(
```
RunningLeon, line 29: may need to pass `device` argument and put the created tensors on the  same device as `pairwise_weight`

RunningLeon (approved):
LGTM

Junjun2016 (approved):

MengzhangLI (approved):

sshuair:
@MeowZheng Could you review this PR, We are waiting for product use. Otherwise we have to modify the code when pip installed the package.

MeowZheng:
sorry for late reply

MeowZheng (approved):

PR #1222, opened 2022-01-20
[Fix] Add `COCOStuffDataset` in `class_names.py`.

MengzhangLI:
## Motivation

Without class names or palette value of `COCOStuffDataset`, community members sometimes may encounter errors when they use COCO-Stuff 10k or COCO-Stuff 164k dataset for visiualization or other operations.

## Modification

Add palette and class names in [`mmseg/core/evaluation/class_names.py`](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/core/evaluation/class_names.py).

### DIFF
--- a/mmseg/core/evaluation/class_names.py
+++ b/mmseg/core/evaluation/class_names.py
@@ -68,6 +68,41 @@ def potsdam_classes():
     ]
 
 
+def cocostuff_classes():
+    """CocoStuff class names for external use."""
+    return [
+        'person', 'bicycle', 'car', 'motorcycle', 'airplane', 'bus', 'train',
+        'truck', 'boat', 'traffic light', 'fire hydrant', 'stop sign',
+        'parking meter', 'bench', 'bird', 'cat', 'dog', 'horse', 'sheep',
+        'cow', 'elephant', 'bear', 'zebra', 'giraffe', 'backpack', 'umbrella',
+        'handbag', 'tie', 'suitcase', 'frisbee', 'skis', 'snowboard',
+        'sports ball', 'kite', 'baseball bat', 'baseball glove', 'skateboard',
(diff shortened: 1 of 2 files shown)


mmseg/core/evaluation/class_names.py
```diff
@@ -140,12 +175,60 @@ def potsdam_palette():
...
+
+
 dataset_aliases = {
     'cityscapes': ['cityscapes'],
     'ade': ['ade', 'ade20k'],
     'voc': ['voc', 'pascal_voc', 'voc12', 'voc12aug'],
     'loveda': ['loveda'],
-    'potsdam': ['potsdam']
+    'potsdam': ['potsdam'],
+    'cocostuff': ['cocostuff']
```
Junjun2016, line 231: Could add more values, e.g. `coco-stuff`, `coco-stuff10k` , `coco-stuff164k`,  `coco_stuff`, '`coco_stuff10k`' , '`coco_stuff164k`'.

MeowZheng: I saw the final version of coco stuff is 164k, so why we add aliases names `coco-stuff10k`?
ref: https://github.com/nightrome/cocostuff#versions-of-coco-stuff

MengzhangLI: It is historically reserved issue: some influential models like BiSeNetV2 (https://arxiv.org/pdf/2004.02147.pdf) use Coco-Stuff 10K and we already provided Coco-Stuff 10K in our configs (https://github.com/open-mmlab/mmsegmentation/tree/master/configs/pspnet#coco-stuff-10k).

MengzhangLI: BTW, same reason (historically reserved benchmark in semantic segmentation) made us keep `Pascal Context` and `Pascal Context 59` simultaneously.

PR #1247, opened 2022-01-27
[Fix] Fix bug in non-distributed multi-gpu training/testing

MengzhangLI:
Related PR:
https://github.com/open-mmlab/mmdetection/pull/7019
https://github.com/open-mmlab/mmaction2/pull/1406
https://github.com/open-mmlab/mmflow/pull/85

`Since MMDP does not support non-distributed multi-GPU training, --gpus in train.py lost its role, so it is removed, --gpu-ids is changed to --gpu-id, because only one GPU can be specified for non-distribution training and testing.
If the number of GPUs is more than 1, there will be an assertion error in MMDP because MMDP does not support the situation with more than one GPU.`

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -35,13 +35,19 @@ def parse_args():
     group_gpus.add_argument(
         '--gpus',
         type=int,
-        help='number of gpus to use '
+        help='(Deprecated, please use --gpu-id) number of gpus to use '
         '(only applicable to non-distributed training)')
     group_gpus.add_argument(
         '--gpu-ids',
         type=int,
         nargs='+',
-        help='ids of gpus to use '
+        help='(Deprecated, please use --gpu-id) ids of gpus to use '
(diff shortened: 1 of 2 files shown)


tools/train.py
```diff
@@ -119,9 +125,14 @@ def main():
     if args.resume_from is not None:
         cfg.resume_from = args.resume_from
     if args.gpu_ids is not None:
-        cfg.gpu_ids = args.gpu_ids
-    else:
-        cfg.gpu_ids = range(1) if args.gpus is None else range(args.gpus)
+        cfg.gpu_ids = args.gpu_ids[0:1]
```
MeowZheng, line 128: add args.gpus 
https://github.com/open-mmlab/mmflow/blob/198748d7856d33154c72f2220a4c44a088b1318e/tools/train.py#L109-L113

MeowZheng (approved):

PR #1251, opened 2022-01-27
[Enhance] New-style CPU training and inference.

MengzhangLI:
## Motivation

Refers to open-mmlab/mmdetection#7016, and it's based on open-mmlab/mmcv#1621

Note: Due to modification of ` mmcv/parallel/_functions.py`, see [pr](https://github.com/open-mmlab/mmcv/pull/1621/files) here.

```python
else:
    # unsqueeze the first dimension thus the tensor's shape is the
    # same as those scattered with GPU.
    output = output.unsqueeze(0)
```

Due to deletion of above lines, so if use CPU function in this pr but mmcv is not the latest version, the error would be raised:

`RuntimeError: Expected 4-dimensional input for 4-dimensional weight [64, 3, 3, 3], but got 5-dimensional input of size [1, 4, 3, 128, 128] instead`

### DIFF
--- a/docs/en/inference.md
+++ b/docs/en/inference.md
@@ -6,6 +6,7 @@ and also some high-level apis for easier integration to other projects.
 ### Test a dataset
 
 - single GPU
+- CPU
 - single node multiple GPU
 - multiple node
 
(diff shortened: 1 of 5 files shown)


MengzhangLI:
In MMSegmentation, it already supports `SyncBN` transferred to `BN` if not DDP.

https://github.com/open-mmlab/mmsegmentation/blob/master/tools/train.py#L176-L182

MengzhangLI:
In this PR, inference could also use CPU environment. 

image

MengzhangLI:
The file conflict would be raised after https://github.com/open-mmlab/mmsegmentation/pull/1247 merged.

Wait for resolve...

MengzhangLI:
After conflicting been solved, the cpu training & testing still works.

image

image

MengzhangLI:
Set `CUDA_VISIBLE_DEVICES=-1` and it works.

image

image

MeowZheng (approved):

PR #1259, opened 2022-01-29
[Fix] Make accuracy take into account ignore_index 

HJoonKwon:
## Motivation

Your repository has been so helpful for my projects and I deeply appreciate it. 
My opinion can be wrong, but please look into it and I hope it can be helpful a bit at least.  

I tried to apply 
``` 
reduce_zero_label=True 
``` 
to ignore the background label, and it turned out that the value of acc_seg from the decode_head was weirdly low while other values such as loss and IoU were all looking good. Looking into the decode_head.py, 
```python
    @force_fp32(apply_to=('seg_logit', ))
    def losses(self, seg_logit, seg_label):
        """Compute segmentation loss."""
        loss = dict()
        seg_logit = resize(
            input=seg_logit,
            size=seg_label.shape[2:],
            mode='bilinear',
            align_corners=self.align_corners)
        if self.sampler is not None:
            seg_weight = self.sampler.sample(seg_logit, seg_label)
        else:
            seg_weight = None
        seg_label = seg_label.squeeze(1)

        if not isinstance(self.loss_decode, nn.ModuleList):
            losses_decode = [self.loss_decode]
        else:
            losses_decode = self.loss_decode
        for loss_decode in losses_decode:
            if loss_decode.loss_name not in loss:
                loss[loss_decode.loss_name] = loss_decode(
                    seg_logit,
                    seg_label,
                    weight=seg_weight,
                    ignore_index=self.ignore_index)
            else:
                loss[loss_decode.loss_name] += loss_decode(
                    seg_logit,
                    seg_label,
                    weight=seg_weight,
                    ignore_index=self.ignore_index)

        loss['acc_seg'] = accuracy(seg_logit, seg_label)
        return loss
```  
the loss_decode receives self.ignore_index as an input argument, while the accuracy doesn't. I think it explains why the accuracy seemed so different from other metrics. Thus, I think the accuracy function should receive the ignore index as an argument as well to take into account the ignore_index to calculate the accuracy of the decode head. 

And, I found that the point_head.py also uses the same function without receiving the ignore_index. So I guess both of them should be fixed.

## Modification

Firstly, the accuracy function can be defined like this
``` 
def accuracy(pred, target, ignore_index=255, topk=1, thresh=None): 
```

and inside of the function, the accuracy can be calculated like this to exclude the ignored index, 
```  
res.append(correct_k.mul_(100.0 / target[target!=ignore_index].numel())) 
```

Secondly, in decode_head.py, we can simply put self.ignore_index into one of the arguments like this, 
``` 
loss['acc_seg'] = accuracy(seg_logit, seg_label, ignore_index=self.ignore_index) 
```

Finally, the same strategy can be applied to point_head.py like this, 
``` 
loss['acc_point'] = accuracy(point_logits, point_label, ignore_index=self.ignore_index)
 ```

## Checklist

### DIFF
--- a/mmseg/models/losses/accuracy.py
+++ b/mmseg/models/losses/accuracy.py
@@ -2,12 +2,13 @@
 import torch.nn as nn
 
 
-def accuracy(pred, target, topk=1, thresh=None):
+def accuracy(pred, target, topk=1, thresh=None, ignore_index=None):
     """Calculate accuracy according to the prediction and target.
 
     Args:
         pred (torch.Tensor): The model prediction, shape (N, num_class, ...)
         target (torch.Tensor): The target of each prediction, shape (N, , ...)
+        ignore_index (int | None): The label index to be ignored. Default: None
         topk (int | tuple[int], optional): If the predictions in ``topk``
             matches the target, the predictions will be regarded as
             correct ones. Defaults to 1.
@@ -43,17 +44,19 @@ def accuracy(pred, target, topk=1, thresh=None):
     if thresh is not None:
         # Only prediction values larger than thresh are counted as correct
         correct = correct & (pred_value > thresh).t()
+    correct = correct[:, target != ignore_index]
     res = []
     for k in topk:
         correct_k = correct[:k].reshape(-1).float().sum(0, keepdim=True)
-        res.append(correct_k.mul_(100.0 / target.numel()))
+        res.append(
+            correct_k.mul_(100.0 / target[target != ignore_index].numel()))
     return res[0] if return_single else res
 
 
 class Accuracy(nn.Module):
     """Accuracy calculation module."""
 
-    def __init__(self, topk=(1, ), thresh=None):
+    def __init__(self, topk=(1, ), thresh=None, ignore_index=None):
         """Module to calculate the accuracy.
 
         Args:
@@ -65,6 +68,7 @@ class Accuracy(nn.Module):
         super().__init__()
         self.topk = topk
         self.thresh = thresh
+        self.ignore_index = ignore_index
 
     def forward(self, pred, target):
         """Forward function to calculate accuracy.
@@ -76,4 +80,5 @@ class Accuracy(nn.Module):
         Returns:
             tuple[float]: The accuracies under different topk criterions.
         """
-        return accuracy(pred, target, self.topk, self.thresh)
+        return accuracy(pred, target, self.topk, self.thresh,
+                        self.ignore_index)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: mmseg/models/losses/accuracy.py [hunk 1]
```diff
@@ -2,12 +2,13 @@
 import torch.nn as nn
 
 
-def accuracy(pred, target, topk=1, thresh=None):
+def accuracy(pred, target, topk=1, thresh=None, ignore_index=255):
```
MeowZheng, line 5: I think it would be better to set `None` as default value of `ignore_index`

mmseg/models/losses/accuracy.py
```diff
@@ -2,12 +2,13 @@
...
     """Calculate accuracy according to the prediction and target.
 
     Args:
         pred (torch.Tensor): The model prediction, shape (N, num_class, ...)
         target (torch.Tensor): The target of each prediction, shape (N, , ...)
+        ignore_index (int | None): The label index to be ignored. Default: 255
```
MeowZheng, line 11:
```suggestion
        ignore_index (int | None): The label index to be ignored. 
            If it is None, any label will be taken into account. Default: None
```

MeowZheng (commented):
please add UT to test this feature in https://github.com/open-mmlab/mmsegmentation/blob/b4314f98c1fd0c4d6d793f115231729461647ae1/tests/test_models/test_losses/test_utils.py#L44

mmseg/models/losses/accuracy.py
HJoonKwon: Sure, I will do it with adding UT by tonight(KST). Thank you for the comment :)

Junjun2016 (commented):
Hi, thanks for your contribution.
Please fix the lint error and add unittests.

HJoonKwon:
I found a bug while making unit tests for it, and it is fixed now. I checked the pre-commit and all passed fine :)

Junjun2016:
> I found a bug while making unit tests for it, and it is fixed now. I checked the pre-commit and all passed fine :)

Great.

HJoonKwon:
Hi! I have a question. Should I do pull-rebase based on the most recent commit on the master branch? or Am I done here?  :)

MeowZheng:
> 

Yes, you should

Junjun2016 (approved):

HJoonKwon:
@MeowZheng Okay, I did pull-rebase based on the current master branch and pushed again!

Issue #1261, opened 2022-01-29
Log twice with mmcv==1.4.4


whu-pzhang:
The terminal text log print twice like this:

![image](https://user-images.githubusercontent.com/8895344/151665582-6a832653-ac52-4783-9cda-b0445ff3f6e6.png)

whu-pzhang:
env info:

```
sys.platform: linux
Python: 3.7.12 | packaged by conda-forge | (default, Oct 26 2021, 06:08:21) [GCC 9.4.0]
CUDA available: True
GPU 0,1: NVIDIA GeForce RTX 2080 Ti
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 10.2, V10.2.89
GCC: gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
PyTorch: 1.7.0
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) oneAPI Math Kernel Library Version 2021.4-Product Build 20210904 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v1.6.0 (Git Hash 5ef631a030a6f73131c77892041042805a06064f)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 10.2
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_61,code=sm_61;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_37,code=compute_37
  - CuDNN 7.6.5
  - Magma 2.5.2
[...]

PR #1267, opened 2022-02-08
[Fix] Fix repeating log by `setup_multi_processes`.

MengzhangLI:
The issue: https://github.com/open-mmlab/mmsegmentation/issues/1261 reports the error.

In `setup_multi_processes`, it uses `get_logger` from MMCV which would initialize new logger if no log initialized. 

If `setup_multi_processes` is used before dump config in `train.py`, it would create two different logs in mmseg.

Solution is simple, just move `setup_multi_processes` behind of `get_logger` part of `train.py`.

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -109,9 +109,6 @@ def main():
     if args.cfg_options is not None:
         cfg.merge_from_dict(args.cfg_options)
 
-    # set multi-process settings
-    setup_multi_processes(cfg)
-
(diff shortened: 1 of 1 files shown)


MeowZheng (approved):

MengzhangLI:
Before modification, the log would repeat `N` times where `N` is number of used GPU.

image

After modification, it is fixed.

image

PR #1271, opened 2022-02-08
[Improve] Add exception for PointRend for support CPU-only usage

xvjiarui:
## Motivation

Support CPU-only usage. User may need to run mmseg on CPU-only. But `mmcv.ops` import will fail on CPU-only installation. Add exception handler to fix that. 

## Modification

Add an exception for PointRend imports

## BC-breaking (Optional)

No


Junjun2016:
Lint error.

Junjun2016 (approved):

PR #1171, opened 2021-12-28
[Feature] Support ISPRS Vaihingen Dataset.

MengzhangLI:
## Motivation

Add geographical dataset Vaihingen, which is related with ISPRS dataset.

Related PR: https://github.com/open-mmlab/mmsegmentation/pull/1097

Due to similar models and its readme files of Potsdam dataset are modified together, this pr should be merged after https://github.com/open-mmlab/mmsegmentation/pull/1097, especially after potential conflict problems are solved.

## To Do List

- [x] Add config files
- [x] Add models & logs
- [x] Add unit test

PR #1292, opened 2022-02-16
[Fix] Fix wrong palette value in vaihingen

MengzhangLI:
## Motivation

In original vaihingen dataset pr: https://github.com/open-mmlab/mmsegmentation/pull/1171.

The value of palette is wrong, which should be the same like [Potsdam](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/datasets/potsdam.py#L17-L18) because those two datasets have same class names.


MengzhangLI:
Note: This PR may have conflict with iSAID dataset pr.

MeowZheng (approved):

PR #1296, opened 2022-02-17
[Enhancement] Add win-ci

MengzhangLI:
## Motivation

Add win-ci in .github/workflows/build.yml

ref pr: https://github.com/open-mmlab/mmediting/pull/732, https://github.com/open-mmlab/mmflow/pull/92.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -242,6 +242,7 @@ jobs:
         run: pip install -e .
       - name: Run unittests
         run: |
+          python -m pip install timm
           coverage run --branch --source mmflow -m pytest tests/
       - name: Generate coverage report
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -214,3 +214,36 @@ jobs:
           env_vars: OS,PYTHON
           name: codecov-umbrella
           fail_ci_if_error: false
+
+  test_windows:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -214,3 +214,37 @@ jobs:
...
+        run: |
+          pip install mmcv-full -f https://download.openmmlab.com/mmcv/dist/cpu/torch1.8/index.html --only-binary mmcv-full
+      - name: Install unittest dependencies
+        run: pip install -r requirements/tests.txt -r requirements/optional.txt
+      - name: Build and install
+        run: pip install -e .
+      - name: Run unittests
+        run: |
+          python -m pip install timm
+          coverage run --branch --source mmflow -m pytest tests/
```
MeowZheng, line 246: mmflow?

MeowZheng (approved):

requirements/runtime.txt
```diff
@@ -1,4 +1,5 @@
 matplotlib
 numpy
+opencv-python>=3
```
MeowZheng, line 3: From mmcv experiments, they don't recommend adding opencv in requirements as opencv have 4 versions and there is a conflict between them

MengzhangLI: Delete `opencv-python>=3` would raise error.

PR #1336, opened 2022-03-01
[Fix] Fix the bug that when all pixels in an image is ignored, the ac…

RockeyCoss:
…curacy calculation raises ZeroDivisionError

## Motivation

Fix the bug that when all pixels in an image is ignored, the accuracy computation raises ZeroDivisionError

## Modification

Fix the bug that when all pixels in an image is ignored, the accuracy computation raises ZeroDivisionError

### DIFF
--- a/mmseg/models/losses/accuracy.py
+++ b/mmseg/models/losses/accuracy.py
@@ -48,8 +48,13 @@ def accuracy(pred, target, topk=1, thresh=None, ignore_index=None):
     res = []
     for k in topk:
         correct_k = correct[:k].reshape(-1).float().sum(0, keepdim=True)
-        res.append(
-            correct_k.mul_(100.0 / target[target != ignore_index].numel()))
+        total_num = target[target != ignore_index].numel()
+        # Avoid causing ZeroDivisionError when all pixels
+        # of an image are ignored
+        if total_num != 0:
+            res.append(correct_k.mul_(100.0 / total_num))
+        else:
+            res.append(100.0)
     return res[0] if return_single else res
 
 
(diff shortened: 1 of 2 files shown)


mmseg/models/losses/accuracy.py
```diff
@@ -48,8 +48,13 @@ def accuracy(pred, target, topk=1, thresh=None, ignore_index=None):
...
         correct_k = correct[:k].reshape(-1).float().sum(0, keepdim=True)
-        res.append(
-            correct_k.mul_(100.0 / target[target != ignore_index].numel()))
+        total_num = target[target != ignore_index].numel()
+        # Avoid causing ZeroDivisionError when all pixels
+        # of an image are ignored
+        if total_num != 0:
+            res.append(correct_k.mul_(100.0 / total_num))
+        else:
+            res.append(100.0)
```
MeowZheng, line 57: `100.` might not be proper.
just check the effect in `mAcc` and mmseg/models/decode_heads/point_head.py

MeowZheng (approved):

MeowZheng (commented):
please fix ut error

MeowZheng (approved):

MengzhangLI (approved):

mmseg/models/losses/accuracy.py
```diff
@@ -46,10 +46,13 @@ def accuracy(pred, target, topk=1, thresh=None, ignore_index=None):
         correct = correct & (pred_value > thresh).t()
     correct = correct[:, target != ignore_index]
     res = []
+    eps = 1e-6
```
Junjun2016, line 49: Should we use eps in `torch.finfo`?
https://pytorch.org/docs/stable/type_info.html?highlight=finfo#torch.torch.finfo

Junjun2016: @MeowZheng

Junjun2016 (approved):
LGTM, except for the definition of `eps`.

mmseg/models/losses/accuracy.py
Junjun2016: Please also refer to https://discuss.pytorch.org/t/torch-epsilon/30497.

PR #1365, opened 2022-03-10
[Enhancement] Add `--return-all` to return logits value for downstream evaluation.

MengzhangLI:
## Motivation

Add `--return-all` to return logits value for downstream evaluation, such as AUC-ROC Curve.

## Update
This pr would be closed, support https://github.com/open-mmlab/mmsegmentation/pull/1374 instead.


mmseg/apis/test.py
```diff
@@ -61,6 +62,9 @@ def single_gpu_test(model,
             Mutually exclusive with pre_eval and efficient_test.
             Default: False.
         format_args (dict): The args for format_results. Default: {}.
+        return_all (bool): Return logits value of prediction.
+            Default: False.
+
```
MeowZheng, line 67: I think it would be better to add a kwargs `metric`,
and send `metric` to `pre_eval`,and `pre_eval` call metric function based on `metric`, but not `pre_eval` just call 'iou'. And for different metric function, just fetch different value from `result`

mmseg/apis/test.py
```diff
@@ -88,7 +92,7 @@ def single_gpu_test(model,
 
     for batch_indices, data in zip(loader_indices, data_loader):
         with torch.no_grad():
-            result = model(return_loss=False, **data)
+            result = model(return_all=return_all, return_loss=False, **data)
```
MeowZheng, line 95: here `return_all=True`

PR #1406, opened 2022-03-23
[Fix] Fix windows-style path in `md2yml.py` in Windows pre-commit.

MengzhangLI:
## Motivation
In Windows OS, the pre-commit would generate `model-index.yml` with Windows path style.

For example:

image

## Solution
Using `.replace('\\', '/')` for path.


MengzhangLI:
I would delete it because it is directly pushed to official repo.

PR #1423, opened 2022-03-28
[fix] fix image save path bug in Windows

whu-pzhang:
## Motivation

when run `.\tools\convert_datasets\isaid.py`  in Windows OS,  as the path str has no `/`，there will get wrong image save path.

image

## Modification

using `os.path.basename(src_path).split('.')[0]` to get the filename without suffix.

### DIFF
--- a/tools/convert_datasets/isaid.py
+++ b/tools/convert_datasets/isaid.py
@@ -85,8 +85,7 @@ def slide_crop_image(src_path, out_dir, mode, patch_H, patch_W, overlap):
 
             img_patch = img[y_str:y_end, x_str:x_end, :]
             img_patch = Image.fromarray(img_patch.astype(np.uint8))
-            image = osp.splitext(
-                src_path.split('/')[-1])[0] + '_' + str(y_str) + '_' + str(
+            image = osp.basename(src_path).split('.')[0] + '_' + str(y_str) + '_' + str(
                     y_end) + '_' + str(x_str) + '_' + str(x_end) + '.png'
             # print(image)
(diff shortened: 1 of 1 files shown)


MengzhangLI:
Hi, thanks for your contribution.

BTW, could this pr run correctly in Linux/Ubuntu OS?

whu-pzhang:
> Hi, thanks for your contribution.
> 
> BTW, could this pr run correctly in Linux/Ubuntu OS?

I runs the script in Ubuntu20.04,  the image filename is current.
image

MengzhangLI:
> 

Pretty cool! So let it run and we can check out whether the number of train/val is identical.

By the way, could you please fix the lint error? Thanks in advance!

Best,

whu-pzhang:
> > 
> 
> Pretty cool! So let it run and we can check out whether the number of train/val is identical.
> 
> By the way, could you please fix the lint error? Thanks in advance!
> 
> Best,

The proceduce is pretty slow, after it finished, I will report the info. 
and the lint error will be fixed in next commit.

whu-pzhang:
> 

The generated train/val images numbers are 33978 and 11644, which is identical with [iSAID](https://github.com/open-mmlab/mmsegmentation/blob/master/docs/en/dataset_prepare.md#isaid)
image

MengzhangLI:
> > 
> 
> The generated train/val images numbers are 33978 and 11644, which is identical with [iSAID](https://github.com/open-mmlab/mmsegmentation/blob/master/docs/en/dataset_prepare.md#isaid) image

`osp.basename` is such a good thing for achieving file name. ;)

MengzhangLI (approved):

PR #1445, opened 2022-04-03
fix bug about label_map

Yzichen:
## Motivation
When I use label_map to change train_label, I found that when the label map is complex, the current code is likely to repeatedly replace gt_semantic_seg. 
For example, when my label_map is dict{6:9, 9:7}, the intention is to replace 6 with 9 and 9 with 7, but the result does replace 6 with 7. 

## Modification
Record the initial gt_semantic_seg with a copy!

### DIFF
--- a/mmseg/datasets/pipelines/loading.py
+++ b/mmseg/datasets/pipelines/loading.py
@@ -135,8 +135,9 @@ class LoadAnnotations(object):
             backend=self.imdecode_backend).squeeze().astype(np.uint8)
         # modify if custom classes
         if results.get('label_map', None) is not None:
+            gt_semantic_seg_copy = gt_semantic_seg.copy()
             for old_id, new_id in results['label_map'].items():
-                gt_semantic_seg[gt_semantic_seg == old_id] = new_id
+                gt_semantic_seg[gt_semantic_seg_copy == old_id] = new_id
         # reduce zero_label
         if self.reduce_zero_label:
             # avoid using underflow conversion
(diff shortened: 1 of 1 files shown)


MeowZheng (approved):
Many thanks for your contribution. There is indeed such a problem that you described.

mmseg/datasets/pipelines/loading.py
```diff
@@ -135,8 +135,9 @@ def __call__(self, results):
             backend=self.imdecode_backend).squeeze().astype(np.uint8)
         # modify if custom classes
         if results.get('label_map', None) is not None:
+            gt_semantic_seg_copy = gt_semantic_seg.copy()
```
MengzhangLI, line 138:
```suggestion
            # Add deep copy to solve bug of repeatedly
            # replace `gt_semantic_seg`, which is reported in 
            # https://github.com/open-mmlab/mmsegmentation/pull/1445/
            gt_semantic_seg_copy = gt_semantic_seg.copy()
```

MengzhangLI: @Yzichen  Hi, thanks for your great pr. I suggest add comment to let users know why we add deep copy and give related pr info.

MeowZheng:
might fix lint problem

MengzhangLI (approved):

lygsbw:
> Many thanks for your contribution. There is indeed such a problem that you described.

There is the same bug in core/evaluation/metrics.py when calculating IoU.

Issue #1430, opened 2022-03-29
Training using Bisenetv2 with custom dataset of 2 classes


JJLimmm:
Hi, if i want to train a model using the bisenetv2 model, and a custom dataset where the classes are either water puddles or background, which files should i be making changes in to ensure that i only want the waterpuddles class to be predicted?

Model: **Bisenetv2**
Dataset: **Custom dataset (ChaseDB1 format)**

Currently, i have made the following changes...

1: Under _configs/_base_/models/bisenetv2.py_, i have changed the number of classes to 2

2: I have created a custom dataset class called _**waterpuddlesDataset()**_, and added into the _mmseg/datasets/waterpuddles.py_ directory. In that file, the classes is defined as **('background', 'waterpuddle')**.

Another question is, how does the **reduce_zero_label** work? 

When i set it to True, my validation accuracy for waterpuddle class become NaN.

JJLimmm:
Sorry, for setting the channel of the output as one, which part of it are you referring to in the base model config file _configs/_base_/models/bisenetv2.py_?

JJLimmm:
> My suggestion is to use BCELoss and set the channel of the output as one.

Sorry, for setting the channel of the output as one, which part of it are you referring to in the base model config file configs/base/models/bisenetv2.py? @MeowZheng

MeowZheng:
num_classes of decode_head

JJLimmm:
> num_classes of decode_head

I changed the channel of the output to 1 and using BCE Loss by setting CrossEntropyLoss's use_sigmoid=True. But i get this error shown below

![error_ss](https://user-images.githubusercontent.com/79782049/160969776-1bef710e-7149-4e62-ba7d-4595b505b453.png)

Below are the config files i used for this run.
Under configs/*
- [configs.zip](https://github.com/open-mmlab/mmsegmentation/files/8385686/configs.zip)


 Under mmseg/datasets/*
- [waterpuddles.zip](https://github.com/open-mmlab/mmsegmentation/files/8385692/waterpuddles.zip)

MeowZheng:
please fetch the master branch, we have fix the bug

JJLimmm:
Hi @MeowZheng , i am still getting the same error after fetching the new fixes. 

![image](https://user-images.githubusercontent.com/79782049/161016301-baaed3bf-5ab3-45e2-bd08-77b70df0e14e.png)

Configs used are still the same as the previous replies above.

[...]

PR #1438, opened 2022-03-31
[Refactory] Merge BEiT and ConvNext 's LR decay optimizer constructors

linfangjian01:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

As BEiT and ConvNext 's LR decay optimizer constructors have similar functions, the difference between them is how to get layer/stage id and we merge them in this pr.

## Modification

1. Rename mmseg/core/utils/layer_decay_optimizer_constructor.py to mmseg/core/optimizers/layer_decay_optimizer_constructor.py 
2. Remove mmseg/core/layer_decay_optimizer_constructor.py
3. Rename `get_num_layer_layer_wise` `get_num_stage_stage_wist`
4. Add a condition for BEiT in `layer_wise`
5. Add deprecated warning to `LayerDecayOptimizerConstructor`
6. Re-organize UTs based on modifications

PR #1454, opened 2022-04-06
[Fix] Support single cahnnel `pred` for Binary Cross Entropy Loss

RockeyCoss:

## Motivation

Fix the bug that binary cross entropy loss doesn't support single channel input.
Related Issue: https://github.com/open-mmlab/mmsegmentation/issues/1430

## Modification

1. When the channel of `pred` is one, instead of expanding the lebel to one-hot format, `binary_cross_entropy_loss` only unsqueeze the `label` to match the shape of the `pred`.
2. Add a new test function to cover the new codes.

### DIFF
--- a/mmseg/models/losses/cross_entropy_loss.py
+++ b/mmseg/models/losses/cross_entropy_loss.py
@@ -135,7 +135,7 @@ def binary_cross_entropy(pred,
         # should mask out the ignored elements
         valid_mask = ((label >= 0) & (label != ignore_index)).float()
         if weight is not None:
-            weight *= valid_mask
+            weight = weight * valid_mask
         else:
             weight = valid_mask
     # average loss over non-ignored and valid elements
--- a/mmseg/models/losses/cross_entropy_loss.py
+++ b/mmseg/models/losses/cross_entropy_loss.py
@@ -120,10 +120,21 @@ def binary_cross_entropy(pred,
                 pred.dim() == 4 and label.dim() == 3), \
             'Only pred shape [N, C], label shape [N] or pred shape [N, C, ' \
             'H, W], label shape [N, H, W] are supported'
-        # `weight` returned from `_expand_onehot_labels`
-        # has been treated for valid (non-ignore) pixels
-        label, weight, valid_mask = _expand_onehot_labels(
-            label, weight, pred.shape, ignore_index)
+
+        if pred.size(1) == 1:
+            # For binary class segmentation, the shape of `pred` is
+            # [N, 1, H, W] and that of `label` is [N, H, W].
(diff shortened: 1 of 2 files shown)


mmseg/models/losses/cross_entropy_loss.py
```diff
@@ -120,10 +120,21 @@ def binary_cross_entropy(pred,
                 pred.dim() == 4 and label.dim() == 3), \
             'Only pred shape [N, C], label shape [N] or pred shape [N, C, ' \
             'H, W], label shape [N, H, W] are supported'
-        # `weight` returned from `_expand_onehot_labels`
-        # has been treated for valid (non-ignore) pixels
-        label, weight, valid_mask = _expand_onehot_labels(
-            label, weight, pred.shape, ignore_index)
+
+        if pred.size(1) == 1:
```
RockeyCoss, line 124: do we need to judge whether `torch.max(label) == 1` here?

MeowZheng (commented):
Please refine pr tittle and pr msg first ref #1438

mmseg/models/losses/cross_entropy_loss.py
```diff
@@ -120,10 +120,21 @@ def binary_cross_entropy(pred,
...
+            # For binary class segmentation, the shape of `pred` is
+            # [N, 1, H, W] and that of `label` is [N, H, W].
+            label = label.unsqueeze(1)
+            valid_mask = ((label >= 0) & (label != ignore_index)).float()
+            if weight is not None:
+                weight = weight.unsqueeze(1).float() * valid_mask
+            else:
+                weight = valid_mask
```
MeowZheng, line 132: I think moving these lines to `_expand_onehot_labels` might be better, as the function of these lines is expanding labels.

mmseg/models/losses/cross_entropy_loss.py
```diff
@@ -65,23 +65,33 @@ def cross_entropy(pred,
...
     """Expand onehot labels to match the size of prediction."""
-    bin_labels = labels.new_zeros(target_shape)
-    valid_mask = (labels >= 0) & (labels != ignore_index)
-    inds = torch.nonzero(valid_mask, as_tuple=True)
-
-    if inds[0].numel() > 0:
-        if labels.dim() == 3:
-            bin_labels[inds[0], labels[valid_mask], inds[1], inds[2]] = 1
+    if target_shape[1] == 1:
+        # For binary class segmentation, the shape of `pred` is
```
MeowZheng, line 69: I think it needs to add assertion like
```python
assert label.max() <= 1, 'balabala'
```

MengzhangLI (approved):

PR #1461, opened 2022-04-09
[Fix] Fix save best checkpoint bug when separate_eval is True

FreyWang:
## Motivation

When use `save_best='mIoU'` or other metric in cfg.evaluation, if `separate_eval=True`, the results metric keys will change to `0_mIoU` and `1_mIoU`... Lead to error when save best checkpoint
## Modification
1. calculate the average metric of datasets if test datasets have same dataset type
e.g. results['mIoU'] = mean(results['0_mIoU], results['1_mIoU]..)
2. disable `save_best` if test datasets have different dataset type and raise warning

### DIFF
--- a/mmseg/datasets/dataset_wrappers.py
+++ b/mmseg/datasets/dataset_wrappers.py
@@ -34,6 +34,8 @@ class ConcatDataset(_ConcatDataset):
         assert separate_eval in [True, False], \
             f'separate_eval can only be True or False,' \
             f'but get {separate_eval}'
+        self.have_same_type = \
+            len(set([type(ds) for ds in self.datasets])) == 1
         if any([isinstance(ds, CityscapesDataset) for ds in datasets]):
             raise NotImplementedError(
                 'Evaluating ConcatDataset containing CityscapesDataset'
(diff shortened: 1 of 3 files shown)


mmseg/datasets/dataset_wrappers.py
```diff
@@ -34,6 +34,8 @@ def __init__(self, datasets, separate_eval=True):
         assert separate_eval in [True, False], \
             f'separate_eval can only be True or False,' \
             f'but get {separate_eval}'
+        self.have_same_type = \
```
MeowZheng, line 37: As this attribution type is bool, so I prefer to `is_same_type` instead of `have_same_type`

FreyWang: well, renamed

mmseg/datasets/dataset_wrappers.py
```diff
@@ -82,10 +85,20 @@ def evaluate(self, results, logger=None, **kwargs):
...
+
+            # calculate the average metric of datasets
+            # if they have same dataset type
+            # e.g. results['mIoU'] = mean(results['0_mIoU], results['1_mIoU]..)
+            if self.have_same_type:
+                for m in metrics:
+                    total_eval_results[m] = np.mean([
+                        total_eval_results[k] for k in total_eval_results
+                        if k.endswith(m)
+                    ])
```
MeowZheng, line 98: I think averaging metrics might be improper, for example, if users would like to split a test dataset into 2 same-type sub-datasets and do evaluation separately, there should be 2 results for sub-datasets and one average result cannot represent model performance. 

I think `save_best` might be turned off when testing datasets separately,

FreyWang: Fine, I will revert and use it only in my private branch😂

FreyWang: reverted

PR #1503, opened 2022-04-23
migrate azure blob for beit checkpoints

donglixp:
## Motivation

We are going to use a new blob account to store the checkpoints.

## Modification

Modify the azure blob storage URLs for BEiT checkpoints.

## BC-breaking

It does not break the compatibility.

### DIFF
--- a/configs/beit/README.md
+++ b/configs/beit/README.md
@@ -46,7 +46,7 @@ python tools/model_converters/beit2mmseg.py ${PRETRAIN_PATH} ${STORE_PATH}
 E.g.
 
 ```shell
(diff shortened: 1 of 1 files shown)


MengzhangLI:
Hi, @donglixp thanks for your nice PR!

@linfangjian01 We may also provide community users with directly pretrained model links, like [swin transformer].(https://github.com/open-mmlab/mmsegmentation/blob/master/configs/swin/upernet_swin_base_patch4_window12_512x512_160k_ade20k_pretrain_384x384_1K.py#L5)

MengzhangLI (approved):

linfangjian01 (approved):

PR #1580, opened 2022-05-14
[Enhancement] Improve Dockerfile of mmsegmentation

atinfinity:
## Motivation

[Current Dockerfile](https://github.com/open-mmlab/mmsegmentation/blob/v0.24.1/docker/Dockerfile) has the following problems.

- There is no README.
- In this docker container, there is no non-root account. So, it may cause permission problems.
- [This Dockerfile](<https://github.com/open-mmlab/mmsegmentation/blob/v0.24.1/docker/Dockerfile#L28>) clone master branch of mmsegmentation.
- Some packages are missing to use mmsegmentation tools.
- In docker container, shared memory size is too small. So, it may cause problems.

## Modification

I changed the following points to resolve mentioned problems.

- I add README to use Dockerfile.
- I changed Dockerfile to create non-root account.
- I changed Dockerfile to specify version of mmsegmentation.
- I changed Dockerfile to install packages for mmsegmentation tools.
  - seaborn : to use `analyze_logs.py`
  - scipy : to use `convert_datasets`
  - sudo : to operate as non-root account
  - curl, wget : to download dataset and pretraind model
  - bash-completion : to enable bash completion in docker container
- I add `--shm-size=4gb` to the option of `docker run`.

## inference

```
mkdir checkpoints
wget https://download.openmmlab.com/mmsegmentation/v0.5/pspnet/pspnet_r50-d8_512x1024_40k_cityscapes/pspnet_r50-d8_512x1024_40k_cityscapes_20200605_003338-2966598c.pth -P checkpoints
python demo/image_demo.py demo/demo.png configs/pspnet/pspnet_r50-d8_512x1024_40k_cityscapes.py checkpoints/pspnet_r50-d8_512x1024_40k_cityscapes_20200605_003338-2966598c.pth --device cuda:0 --palette cityscapes
```

## training

```
python tools/train.py configs/pspnet/pspnet_r50-d8_512x512_20k_voc12aug.py
```

## Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,7 @@
 ARG PYTORCH="1.11.0"
 ARG CUDA="11.3"
 ARG CUDNN="8"
+ARG MMCV="1.5.0"
 
 FROM pytorch/pytorch:${PYTORCH}-cuda${CUDA}-cudnn${CUDNN}-devel
 
@@ -40,7 +41,7 @@ SHELL ["/bin/bash", "-c"]
 ARG PYTORCH
 ARG CUDA
 ARG MMCV
-RUN pip install --no-cache-dir mmcv-full -f https://download.openmmlab.com/mmcv/dist/cu${CUDA//./}/torch${PYTORCH}/index.html
+RUN pip install --no-cache-dir mmcv-full==${MMCV} -f https://download.openmmlab.com/mmcv/dist/cu${CUDA//./}/torch${PYTORCH}/index.html
 
 # Install MMSegmentation
 ARG MMSEG_VERISON=master
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,6 +2,7 @@ ARG PYTORCH="1.11.0"
 ARG CUDA="11.3"
 ARG CUDNN="8"
 ARG MMCV="1.5.0"
+ARG MMSEG="master"
 
 FROM pytorch/pytorch:${PYTORCH}-cuda${CUDA}-cudnn${CUDNN}-devel
 
@@ -44,8 +45,8 @@ ARG MMCV
 RUN pip install --no-cache-dir mmcv-full==${MMCV} -f https://download.openmmlab.com/mmcv/dist/cu${CUDA//./}/torch${PYTORCH}/index.html
 
 # Install MMSegmentation
-ARG MMSEG_VERISON=master
-RUN git clone https://github.com/open-mmlab/mmsegmentation.git /home/$USERNAME/mmsegmentation -b ${MMSEG_VERISON}
+ARG MMSEG
+RUN git clone https://github.com/open-mmlab/mmsegmentation.git /home/$USERNAME/mmsegmentation -b ${MMSEG}
 WORKDIR /home/$USERNAME/mmsegmentation
 ENV FORCE_CUDA="1"
 RUN pip install -r requirements.txt
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/machi
 ARG GID=1000
 ARG UID=1000
 ENV USERNAME mmseg
-ENV HOME /home/$USERNAME 
+ENV HOME /home/$USERNAME
 RUN groupadd -f -g ${GID} ${USERNAME}
 RUN useradd -m $USERNAME -u ${UID} -g ${GID} && \
     echo "$USERNAME:$USERNAME" | chpasswd && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,21 +12,46 @@ ENV CMAKE_PREFIX_PATH="$(dirname $(which conda))/../"
 RUN apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/3bf863cc.pub
 RUN apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1804/x86_64/7fa2af80.pub
 
-RUN apt-get update && apt-get install -y git ninja-build libglib2.0-0 libsm6 libxrender-dev libxext6 libgl1-mesa-glx \
+# add new sudo user
+ARG GID=1000
+ARG UID=1000
+ENV USERNAME mmseg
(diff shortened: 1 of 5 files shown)


MengzhangLI:
Hi, @atinfinity thanks for your nice PR!

We would review it as soon as poosible.

Best,

docker/Dockerfile
```diff
@@ -12,21 +12,46 @@ ENV CMAKE_PREFIX_PATH="$(dirname $(which conda))/../"
 RUN apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/3bf863cc.pub
 RUN apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1804/x86_64/7fa2af80.pub
 
-RUN apt-get update && apt-get install -y git ninja-build libglib2.0-0 libsm6 libxrender-dev libxext6 libgl1-mesa-glx \
+# add new sudo user
+ARG GID=1000
+ARG UID=1000
+ENV USERNAME mmseg
+ENV HOME /home/$USERNAME
+RUN groupadd -f -g ${GID} ${USERNAME}
```
MeowZheng, line 20: I think it might be `$USERNAME`, as USERNAME is a ENV variables

docker/launch_container.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+if [ $# != 1 ]; then
+    echo "usage: $0 <mmseg_version>"
+    exit 1
+fi
+
+mmseg_version=$1
+
+docker run --gpus all -it \
```
MeowZheng, line 10: might emphasize docker >=19.03 in README

MeowZheng (commented):
Could you please let me know what the usage scenario is, as some users might not need to add non-root accounts and x11?

atinfinity:
@MeowZheng

> Could you please let me know what the usage scenario is, as some users might not need to add non-root accounts and x11?

These are different problems.

Official [demo scripts](https://github.com/open-mmlab/mmsegmentation/tree/v0.24.1/demo) need X11. So, I enabled X11 in docker container. But, these scripts don't necessarily require non-root users.

On the other hands, mmsegmentation has useful convert tools.
When root user use this tool in docker container, ownership of converted data becomes root user.
So, it may cause permission problems outside of docker container.

MeowZheng:
> @MeowZheng
> 
> > Could you please let me know what the usage scenario is, as some users might not need to add non-root accounts and x11?
> 
> These are different problems.
> 
> Official [demo scripts](https://github.com/open-mmlab/mmsegmentation/tree/v0.24.1/demo) need X11. So, I enabled X11 in docker container. But, these scripts don't necessarily require non-root users.
> 
> On the other hands, mmsegmentation has useful convert tools. When root user use this tool in docker container, ownership of converted data becomes root user. So, it may cause permission problems outside of docker container.

It seems create user accounts within docker images is not a common practice. Most images pulled from docker hub has only the root user. 
How about passing `-u` arguments to `docker run` to solve the permission issue, like [this post](https://stackoverflow.com/questions/71501506/userid-and-group-id-in-docker-run-command) suggests?

docker/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+./launch_container.sh 0.24.1
+```
+
+### with X11
+
+```
+./launch_container_x11.sh <mmseg_version>
+```
+
```
MeowZheng, line 40:
```suggestion
Official [demo scripts](https://github.com/open-mmlab/mmsegmentation/tree/v0.24.1/demo) need X11. So, if you would like to visualize the results, please run the script `launch_container_x11.sh` to enable X11 in docker container.
```

atinfinity: :+1:

atinfinity:
@MeowZheng 

> It seems create user accounts within docker images is not a common practice. Most images pulled from docker hub has only the root user.
How about passing -u arguments to docker run to solve the permission issue, like [this post](https://stackoverflow.com/questions/71501506/userid-and-group-id-in-docker-run-command) suggests?

Thank you for your suggestion. I tried this. But, this way has problems.

### Host

```bash
$ docker run -it --rm -u $(id -u):$(id -g) ubuntu:20.04 bash
groups: cannot find name for group ID 1000
I have no name!
```

It occurs the following error.

> groups: cannot find name for group ID 1000
> I have no name!

### Docker container

It occurs the following error.

```bash
$ groups
groups: cannot find name for group ID 1000
1000
```

```bash
$ sudo apt update
bash: sudo: command not found
```

So, user could not install package using apt in docker container.

atinfinity:
I found that current Dockerfile does not specify MMCV. So, it may cause problem of compatible.
I add to default version in <https://github.com/open-mmlab/mmsegmentation/pull/1580/commits/f9d057d2848dba042bc0d6b7beed2a9d8a2a5aab>.

xiexinch:
> @MeowZheng
> 
> > Could you please let me know what the usage scenario is, as some users might not need to add non-root accounts and x11?
> 
> These are different problems.
> 
> Official [demo scripts](https://github.com/open-mmlab/mmsegmentation/tree/v0.24.1/demo) need X11. So, I enabled X11 in docker container. But, these scripts don't necessarily require non-root users.
> 
> On the other hands, mmsegmentation has useful convert tools. When root user use this tool in docker container, ownership of converted data becomes root user. So, it may cause permission problems outside of docker container.

Hi @atinfinity , 
How about modify the access permission inside docker container? That might be a solution to the permission problems.

atinfinity:
@xiexinch 

> How about modify the access permission inside docker container? That might be a solution to the permission problems.

Thank you for your suggestion. This method may solve permission problems.
On the other hands, I think that the following issues remain.
<https://github.com/open-mmlab/mmsegmentation/pull/1580#issuecomment-1132964820>

@MeowZheng If there is no progress, I will close this PR in the near future.

MeowZheng:
Many thanks again for your nice PR. The dockerfile in mmseg just provides the basic and general function, and If you have customized requirements, just modified it locally, which may not be suitable to merge in master.

atinfinity:
@MeowZheng I understood. In official document, there is the following description.

> If a service can run without privileges, use USER to change to a non-root user. Start by creating the user and group in the Dockerfile with something like RUN groupadd -r postgres && useradd --no-log-init -r -g postgres postgres.

<https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user>

PR #1704, opened 2022-06-24
fix OpenCV num_threads log

Sanster:
Correctly print the num_threads of opencv


MeowZheng:
many thanks for your nice pr, please sign CLA first

MeowZheng (approved):

PR #1707, opened 2022-06-25
[Fix] Add notation of training setting of SegFormer

MengzhangLI:
Add notation in README that we adopt default training pipeline in MMSegmentation while SegFormer original repo uses `RepeatDataset` for training acceleration.

### DIFF
--- a/configs/segformer/README.md
+++ b/configs/segformer/README.md
@@ -124,4 +124,4 @@ if args.aug_test:
 
 - Training of SegFormer is not very stable, which is sensitive to random seeds.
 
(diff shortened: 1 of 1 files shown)


configs/segformer/README.md
```diff
@@ -123,3 +123,5 @@ if args.aug_test:
 ```
 
 - Training of SegFormer is not very stable, which is sensitive to random seeds.
+
+- We adopt default training setting in MMSegmentation while original SegFormer repo uses `RepeatDataset` to accelerate [training](https://github.com/NVlabs/SegFormer/blob/master/local_configs/_base_/datasets/ade20k_repeat.py#L38-L39), here is its related [issue](https://github.com/NVlabs/SegFormer/issues/25).
```
MeowZheng, line 127: original SegFormer repo -> SegFormer official repo

MeowZheng, line 127: while -> rather than

PR #1718, opened 2022-06-27
issues about the code

bowenroom:
Hi! Thanks for releasing the code. I re-implement the code and found that the result is not as good as what the paper has demonstrated. There is a gap between two results especially the car. I follows the step Readme has pointed out. Can you release the data used in the experiment? (data in baidu yun are the not sliced ones, I think they are not the data used in the experiment)


bowenroom:
sorry

Issue #1740, opened 2022-07-06
Remove *show=False* argument in WandB Hook  


TimoK93:
Hello everyone,

i noticed a little bug which causes an exception if logging is performed with WandB Hook during training. In file 
https://github.com/open-mmlab/mmsegmentation/blob/5cb9f43c9d1c802d84ab8fa420d1f0bd5ae93177/mmseg/core/hook/wandblogger_hook.py#L206-L207
the argument ``show=False`` is used. The member `self.test_fn` can either be  `single_gpu_test` or `multi_gpu_test` as shown in the following code snippet.
https://github.com/open-mmlab/mmsegmentation/blob/5cb9f43c9d1c802d84ab8fa420d1f0bd5ae93177/mmseg/core/hook/wandblogger_hook.py#L109-L119
The argument `show=False` is not existing in `multi_gpu_test` and set to false per default in `single_gpu_test`. So the argument is not required and could be removed. This would avoid the above shown exception.

**Exception Output**
```
Traceback (most recent call last):
  File "tools/train.py", line 242, in <module>
    main()
  File "tools/train.py", line 231, in main
    train_segmentor(
  File "/home/kaiser/code_dev/mmsegmentation/mmseg/apis/train.py", line 194, in train_segmentor
    runner.run(data_loaders, cfg.workflow)
  File "/home/kaiser/anaconda3/tmp/envs/openmmlab/lib/python3.8/site-packages/mmcv/runner/iter_based_runner.py", line 138, in run
    iter_runner(iter_loaders[i], **kwargs)
  File "/home/kaiser/anaconda3/tmp/envs/openmmlab/lib/python3.8/site-packages/mmcv/runner/iter_based_runner.py", line 68, in train
    self.call_hook('after_train_iter')
  File "/home/kaiser/anaconda3/tmp/envs/openmmlab/lib/python3.8/site-packages/mmcv/runner/base_runner.py", line 309, in call_hook
    getattr(hook, fn_name)(self)
  File "/home/kaiser/anaconda3/tmp/envs/openmmlab/lib/python3.8/site-packages/mmcv/runner/hooks/evaluation.py", line 259, in after_train_iter
    hook.after_train_iter(runner)
  File "/home/kaiser/anaconda3/tmp/envs/openmmlab/lib/python3.8/site-packages/mmcv/runner/dist_utils.py", line 135, in wrapper
    return func(*args, **kwargs)
[...]

PR #1741, opened 2022-07-06
[Fix] Remove `show=False` argument in WandbHook

xiexinch:
## Motivation

#1740 

## Modification

Remove unnecessary `show=False` argument in test_fn


ayulockin:
Thanks for opening this.

PR #1748, opened 2022-07-08
Add Semantic Segmentation of KITTI-STEP Panoptic Segmentation Dataset

TimoK93:
## Motivation

In the last years, panoptic segmentation has become more into the focus in reseach. Weber et al. [[Link]](http://www.cvlibs.net/publications/Weber2021NEURIPSDATA.pdf) have published a quite nice dataset, which is in the same style like Cityscapes, but for KITTI sequences. Since Cityscapes and KITTI-STEP share the same classes and also a comparable domain (dashcam view), interesting investigations, e.g. about relations in the domain e.t.c. can be done. 

Note that KITTI-STEP provices panoptic segmentation annotations which are out of scope for mmsegmentation.

## Modification

Mostly, I added the new dataset and dataset preparation file. To simplify the first usage of the new dataset, I also added configs for the dataset and deeplabv3plus.  

## BC-breaking (Optional)

No BC-breaking

## Use cases (Optional)

Researchers want to test their new methods, e.g. for interpretable AI in the context of semantic segmentation. They want to show, that their method is reproducible on comparable datasets. Thus, they can compare Cityscapes and KITTI-STEP.

### DIFF
--- a/docs/zh_cn/get_started.md
+++ b/docs/zh_cn/get_started.md
@@ -153,11 +153,11 @@ pip install mmcv-full -f https://download.openmmlab.com/mmcv/dist/cu113/torch1.1
 
 ### 在 CPU 环境中安装
 
-MMPose 可以仅在 CPU 环境中安装，在 CPU 模式下，您可以完成训练（需要 MMCV 版本 >= 1.4.4）、测试和模型推理等所有操作。
+MMSegmentation 可以仅在 CPU 环境中安装，在 CPU 模式下，您可以完成训练（需要 MMCV 版本 >= 1.4.4）、测试和模型推理等所有操作。
 
 ### 在 Google Colab 中安装
 
-[Google Colab](https://colab.research.google.com/) 通常已经包含了 PyTorch 环境，因此我们只需要安装 MMCV 和 MMPose 即可，命令如下：
+[Google Colab](https://colab.research.google.com/) 通常已经包含了 PyTorch 环境，因此我们只需要安装 MMCV 和 MMSegmentation 即可，命令如下：
 
 **第一步** 使用 [MIM](https://github.com/open-mmlab/mim) 安装 [MMCV](https://github.com/open-mmlab/mmcv)
 


MengzhangLI:
Hi, very thankful for your nice PR. We would review it ASAP.

configs/_base_/datasets/kittistep_368x368.py
```diff
@@ -0,0 +1,35 @@
+_base_ = './kittistep.py'
```
MeowZheng, line 1: Sorry, I didn't see any difference between the `kittistep.py` and `kittistep_368x368.py`
Would you like to tell us the difference?

docs/en/dataset_prepare.md
```diff
@@ -376,3 +380,33 @@ python tools/convert_datasets/isaid.py /path/to/iSAID
 ```
 
 In our default setting (`patch_width`=896, `patch_height`=896,　`overlap_area`=384), it will generate 33978 images for training and 11644 images for validation.
+
+### KITTI-STEP
+
+After registration, the data images could be download from [KITTI-STEP](http://www.cvlibs.net/datasets/kitti/eval_step.php)
+
+You may need to follow the following structure for dataset preparation after downloading iSAID dataset.
```
MeowZheng, line 388: iSAID -> KITTI-STEP

configs/_base_/datasets/kittistep.py
```diff
@@ -0,0 +1,54 @@
...
+        ])
+]
+data = dict(
+    samples_per_gpu=2,
+    workers_per_gpu=2,
+    train=dict(
+        type=dataset_type,
+        data_root=data_root,
+        img_dir='training_openmmlab/image_02/train',
+        ann_dir='panoptic_maps_openmmlab/train',
```
MeowZheng, line 41: I have a little question, is KITTI-STEP for panoptic segmentation tasks or semantic segmentation? MMSeg only works on semantic segmentation, and panoptic segmentation and instance segmentation are supported in MMDet

configs/_base_/datasets/kittistep_368x368.py
TimoK93: Yes you are right... There is no difference. ...368x368.py could be deleted!

docs/en/dataset_prepare.md
TimoK93: Oh, sorry. Sure!

configs/_base_/datasets/kittistep.py
TimoK93: KITTI-STEP is for panoptic segmentation. But because it follows the Cityscapes label set and the same camera setting (car dashcam view), it is a beautiful dataset to compare sem-seg methods especially if they address ML-interpretatbility. E.g., with this dataset you an compare interclass dependencies or interclass confusion on similar datasets.

MeowZheng: I see. Thanks for your explanation.
Is there any public research work using it as a sem-seg dataset, since the baseline result in MMSeg needs a public reference?

TimoK93: Thanks for your effort. The closest baseline can be seen here: http://www.cvlibs.net/publications/Weber2021NEURIPSDATA.pdf in the appendix G. They report STQ metric which is based on SQ(=mIoU). So far, there is no published work using it as a pure Sem-seg dataset. But i am working on it so that i hope it will in some months :) If you think it is more convenient, we also could wait some months and I can create a new PR after publication.

MeowZheng: Of course, we really appreciate it that users release their work in mmseg!

TimoK93: So should i wait with the dataset PR until i used it in a publication?

PR #1759, opened 2022-07-12
Keep training and validation pipeline separated

mawanda-jun:
When structuring the configuration, it is useful to keep training and validation pipelines separated (in case of heavy augmentations vs original validation dataset).

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -214,7 +214,7 @@ def main():
     datasets = [build_dataset(cfg.data.train)]
     if len(cfg.workflow) == 2:
         val_dataset = copy.deepcopy(cfg.data.val)
-        val_dataset.pipeline = cfg.data.train.pipeline
+        val_dataset.pipeline = cfg.data.val.pipeline
         datasets.append(build_dataset(val_dataset))
     if cfg.checkpoint_config is not None:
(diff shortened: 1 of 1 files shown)


tools/train.py
```diff
@@ -214,7 +214,7 @@ def main():
     datasets = [build_dataset(cfg.data.train)]
     if len(cfg.workflow) == 2:
         val_dataset = copy.deepcopy(cfg.data.val)
-        val_dataset.pipeline = cfg.data.train.pipeline
+        val_dataset.pipeline = cfg.data.val.pipeline
```
MeowZheng, line 217: validation workflow is for validation loss calculation, and usually the data augmentation is same as training
https://github.com/open-mmlab/mmsegmentation/blob/fe16402251bbd80092f482c23857e4c6a8694320/mmseg/models/segmentors/base.py#L148-L168

mawanda-jun: I see. I thought it'd give more flexibility to the end user to keep pipelines separated - since, if the same pipeline should be used, it should just be put equal in the configuration file. In my case, it'd be more suitable to have them separated as the two datasets have not the same distribution.

PR #1783, opened 2022-07-15
Added support for albumentations

mawanda-jun:
## Motivation

Add Albumentations support to the mmsegmentation framework

## Modification

I took inspiration from mmdet and made it compatible with the `mask` key (instead of `masks` and `bboxes`). I've removed everything connected with instance segmentation masks and bboxes.

The only concern is: albumentations works with RGB images, while mmcv reads them in BGR format by default. I've added the explicit conversion at lines 1455 and 1460, but I've seen that in mmdet this is not taken in consideration. Is the conversion made somewhere else in that case?

## BC-breaking (Optional)

Should not.

## Use cases (Optional)

The same listed in mmdet

## Checklist

### DIFF
--- a/mmseg/datasets/pipelines/transforms.py
+++ b/mmseg/datasets/pipelines/transforms.py
@@ -1338,6 +1338,7 @@ class RandomMosaic(object):
         repr_str += f'seg_pad_val={self.pad_val})'
         return repr_str
 
+
 @PIPELINES.register_module()
 class Albu:
     """Albumentation augmentation.
--- a/mmseg/datasets/pipelines/transforms.py
+++ b/mmseg/datasets/pipelines/transforms.py
@@ -5,6 +5,10 @@ import mmcv
 import numpy as np
 from mmcv.utils import deprecated_api_warning, is_tuple_of
 from numpy import random
+import cv2
+import albumentations
+from albumentations import Compose
+import inspect
 
 from ..builder import PIPELINES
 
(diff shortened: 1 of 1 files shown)


mmseg/datasets/pipelines/transforms.py
```diff
@@ -5,6 +5,10 @@
 import numpy as np
 from mmcv.utils import deprecated_api_warning, is_tuple_of
 from numpy import random
+import cv2
+import albumentations
+from albumentations import Compose
```
MeowZheng, line 10: As `albumentations` is not tightly coupled with mmseg, I think it would be better that mmseg can be installed without `albumentations`
```suggestion
try:
    import albumentations
    from albumentations import Compose
except ImportError:
    albumentations = None
    Compose = None
```

MeowZheng (commented):
As there is no albumentations in CI (Continuous integration) environment of mmseg, all checks failed, so it is necessary to install albumentation in CI test environment.

ref: 
https://github.com/open-mmlab/mmdetection/blob/56e42e72cdf516bebb676e586f408b98f854d84c/.github/workflows/build.yml#L74

https://github.com/open-mmlab/mmdetection/blob/56e42e72cdf516bebb676e586f408b98f854d84c/.github/workflows/build.yml#L140

xiexinch:
https://github.com/open-mmlab/mmsegmentation/pull/2710

PR #1792, opened 2022-07-18
[Feature] Add SegVisualizer

MengzhangLI:

PR #1794, opened 2022-07-18
[Fix] Fix demo scripts

xiexinch:
## Motivation

Fix demo scripts

## Modification

- apply 2.0 methods to inference APIs
- Refactored the `inference_model()` API

## Dependency

#1792

### DIFF
--- a/mmseg/datasets/__init__.py
+++ b/mmseg/datasets/__init__.py
@@ -15,7 +15,8 @@ from .night_driving import NightDrivingDataset
 from .pascal_context import PascalContextDataset, PascalContextDataset59
 from .potsdam import PotsdamDataset
 from .stare import STAREDataset
-from .transforms import (CLAHE, AdjustGamma, LoadAnnotations, PackSegInputs,
+from .transforms import (CLAHE, AdjustGamma, LoadAnnotations,
+                         LoadImageFromNDArray, PackSegInputs,
                          PhotoMetricDistortion, RandomCrop, RandomCutOut,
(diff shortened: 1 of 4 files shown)


mmseg/apis/inference.py
```diff
@@ -1,12 +1,16 @@
...
-import matplotlib.pyplot as plt
+from typing import Union
+
 import mmcv
+import numpy as np
 import torch
-from mmcv.parallel import collate, scatter
 from mmcv.runner import load_checkpoint
 
 from mmseg.datasets.transforms import Compose
```
MeowZheng, line 9: from mmseg.datasets.transforms.compose import Compose

mmseg/apis/inference.py
```diff
@@ -40,34 +44,50 @@ def init_model(config, checkpoint=None, device='cuda:0'):
...
+    cfg = model.cfg
+    if dict(type='LoadAnnotations') in cfg.test_pipeline:
+        cfg.test_pipeline.remove(dict(type='LoadAnnotations'))
+    # TODO: Consider using the singleton pattern to avoid building
+    # a pipeline for each inference
+    pipeline = Compose(cfg.test_pipeline)
+
+    if isinstance(img, str):
+        data, data_samples = model.data_preprocessor(
+            [pipeline(dict(img_path=img))], False)
```
MeowZheng, line 58: can we use `test_step` of model which can call `data_preprocessor` automatically 
https://github.com/open-mmlab/mmdetection/blob/a0a0211217edf66cd66b8e4074e45cc0f3dafbf7/mmdet/apis/inference.py#L137

PR #1798, opened 2022-07-19
[Enhancement] Replace MMCV with MMEngine in convert model scripts

xiexinch:
## Modification

- Replace MMCV with MMEngine


mmseg/utils/set_env.py
```diff
@@ -18,6 +18,7 @@ def register_all_modules(init_default_scope: bool = True) -> None:
     """  # noqa
     import mmseg.data  # noqa: F401,F403
     import mmseg.datasets  # noqa: F401,F403
+    import mmseg.datasets.transforms  # noqa: F401,F403
```
MeowZheng, line 21:
```suggestion
```

PR #1810, opened 2022-07-22
[Feature] Add LoadImageFromNdArray pipeline

xiexinch:
## Motivation

Can be used when loading images from the webcam.

## Modification

- add LoadImageFromNDArray class
- add unit test

Ref: https://github.com/open-mmlab/mmdetection/blob/a0a0211217edf66cd66b8e4074e45cc0f3dafbf7/mmdet/datasets/transforms/loading.py#L16

### DIFF
--- a/mmseg/datasets/transforms/__init__.py
+++ b/mmseg/datasets/transforms/__init__.py
@@ -1,6 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 from .formatting import PackSegInputs
-from .loading import LoadAnnotations
+from .loading import LoadAnnotations, LoadImageFromNDArray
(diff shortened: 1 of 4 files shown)


mmseg/datasets/transforms/__init__.py
```diff
@@ -1,12 +1,13 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 from .formatting import PackSegInputs
-from .loading import LoadAnnotations
+from .loading import LoadAnnotations, LoadImageFromNDArray
```
MeowZheng, line 3: might add `LoadImageFromNDArray` in `mmseg/datasets/__init__.py`

PR #1829, opened 2022-07-26
[Fix] Use MMEngine BaseModule

MengzhangLI:
## Motivation

Import `BaseModule` from MMEngine rather than MMCV.


mmseg/models/necks/fpn.py
```diff
@@ -2,7 +2,8 @@
 import torch.nn as nn
 import torch.nn.functional as F
 from mmcv.cnn import ConvModule
-from mmcv.runner import BaseModule, auto_fp16
+from mmcv.runner import auto_fp16
```
MeowZheng, line 5:
```suggestion
```

PR #1994, opened 2022-08-30
[Fix] Fix several config file errors in 2.0 (New)

MengzhangLI:
## Motivation
(1) Fix several wrong config files in 2.0.
### Some errors
- Use `init_cfg` and `pretrained` at the same time.
- In config file, `[]` can not be parsed like `{}` so the content in `[]` should be all set.  

(2) Add several config files which are not included at the first time, such as UperNet-R18 and Swin-Large.

## Notification

Those config files are all checked locally.

## Note
Old PR: https://github.com/open-mmlab/mmsegmentation/pull/1886


xiexinch (commented):
Should we modify the readme according to the new config?

MeowZheng (approved):

PR #2075, opened 2022-09-15
fix bug about label_map in evaluation part

rlatjcj:
## Motivation

According to https://github.com/open-mmlab/mmsegmentation/pull/1445, evaluation part should be changed, too.

## Modification

Record the initial label with `.clone()`, because `label` is `torch.tensor`.


MeowZheng (approved):
Thanks for the bug fix, it looks good to me. Please sign CLA first

PR #2088, opened 2022-09-19
[Doc] Update transforms Doc

MengzhangLI:


docs/en/advanced_guides/transforms.md
```diff
@@ -43,47 +69,104 @@ test_pipeline = [
...
+
+```python
+pipeline = [
+    ...
+    dict(type='MyFlip', direction='horizontal'),
+    ...
+]
+```
+
+Note that if you want to use `MyFlip` in config, you must ensure the file containing `MyFlip` is imported during the program run.
```
xiexinch, line 172: I think it could be modified to "during the runtime"

MeowZheng (approved):

PR #2089, opened 2022-09-19
[Fix] Fix MMSegmentation colab demo

MeowZheng:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


demo/MMSegmentation_Tutorial.ipynb
```diff
@@ -68,8 +68,12 @@
    "source": [
     "# Install PyTorch\n",
     "!conda install pytorch=1.10.0 torchvision cudatoolkit=11.1 -c pytorch\n",
+    "# Install mmengine\n",
+    "!pip install git+https://github.com/open-mmlab/mmengine.git@main\n",
```
xiexinch, line 72: I think it would be better to lead the users to install MMEngine with mim

MeowZheng: as mim depends on mmengine, mmengine must be installed before mim

mmseg/apis/inference.py
```diff
@@ -19,10 +19,10 @@
 from mmseg.visualization import SegLocalVisualizer
 
 
-def init_model(config: Union[str, Path, Config],
-               checkpoint: Optional[str] = None,
-               device: str = 'cuda:0',
-               cfg_options: Optional[dict] = None):
+def init_mfodel(config: Union[str, Path, Config],
```
xiexinch, line 22:
```suggestion
def init_model(config: Union[str, Path, Config],
```

xiexinch (approved):

PR #2093, opened 2022-09-20
[Doc] Add Engine Doc

MengzhangLI:


docs/en/advanced_guides/engine.md
```diff
@@ -1 +1,260 @@
 # Engine
+
+<!-- TOC -->
```
xiexinch, line 3: I think we could briefly introduce the `runner`.

xiexinch: Or add a link to the runner documentation. https://mmengine.readthedocs.io/en/latest/tutorials/runner.html

PR #2185, opened 2022-10-13
[Fix] Fix ResizeToMultiple transform in MMSeg 1.x

MengzhangLI:
## Motivation 
Fix ResizeToMultiple transform in MMSeg 1.x.

## Description
In current  `ResizeToMultiple`, it has:
```python
        results['img'] = img
        results['img_shape'] = img.shape
        results['pad_shape'] = img.shape
```

If `dict(type='ResizeToMultiple', size_divisor=32),` is added, because in https://github.com/open-mmlab/mmsegmentation/blob/dev-1.x/mmseg/models/utils/wrappers.py#L27, the `input` and  `size` should have same dimension, but currently it is 2 and 3, respectively. And it would cause error `ValueError: size shape must match input shape. Input is 2D, size is 3`:

image

## Solution
Just modify `ResizeToMultiple`:
```python
        results['img'] = img
        results['img_shape'] = img.shape[:2]
        results['pad_shape'] = img.shape[:2]
```
image

### DIFF
--- a/configs/segformer/README.md
+++ b/configs/segformer/README.md
@@ -77,20 +77,13 @@ using `AlignedResize`, you can change the dataset pipeline like this:
 ```python
 test_pipeline = [
     dict(type='LoadImageFromFile'),
-    dict(
-        type='MultiScaleFlipAug',
-        img_scale=(2048, 512),
-        # img_ratios=[0.5, 0.75, 1.0, 1.25, 1.5, 1.75],
-        flip=False,
-        transforms=[
-            dict(type='Resize', keep_ratio=True),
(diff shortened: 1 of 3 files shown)


MeowZheng (approved):

ThorPham:
I has same error like this when i train customer dataset. But i can't fix with modify 
`        results['img'] = img
        results['img_shape'] = img.shape[:2]
        results['pad_shape'] = img.shape[:2]`
I printed function resize in wrappers.py
I don't know why the last is shape Size([512, 512, 3])  
image

  Here is log error
image

PR #2321, opened 2022-11-18
optim custom dataset

Bin-ze:

## Motivation

This improvement allows to find the corresponding image file through gt file when reading data, so that the image and label of the split data set are no longer strictly aligned, but the label is a subset of the image, which simplifies to a certain extent The data processing steps for training custom dataset are described.

## Modification

~/mmseg/datasets/custom.py

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmseg/datasets/custom.py
+++ b/mmseg/datasets/custom.py
@@ -164,13 +164,14 @@ class CustomDataset(Dataset):
                 img_infos.append(img_info)
         else:
             for img in self.file_client.list_dir_or_file(
-                    dir_path=img_dir,
+                    dir_path=ann_dir,
                     list_dir=False,
-                    suffix=img_suffix,
+                    suffix=seg_map_suffix,
                     recursive=True):
-                img_info = dict(filename=img)
+                img_info = dict(
+                    filename=img.replace(seg_map_suffix, img_suffix))
                 if ann_dir is not None:
-                    seg_map = img.replace(img_suffix, seg_map_suffix)
+                    seg_map = img
                     img_info['ann'] = dict(seg_map=seg_map)
                 img_infos.append(img_info)
             img_infos = sorted(img_infos, key=lambda x: x['filename'])


MeowZheng (commented):
Many thanks for your contribution. Would you like to provide an example of a use case for this modification?

Bin-ze:
> Many thanks for your contribution. Would you like to provide an example of a use case for this modification?

# motivation
When the image and the label are not in one-to-one correspondence, the previous method fails. This PR allows the label to be a subset of the image, go to the image directory to get the corresponding image through the label, instead of finding the label through the image.

# example 
custom datset
| - image
| - |- image1.jpg
| - |- image2.jpg
| - |- image3.jpg
| - |- image4.jpg
| - annotations
| - |- image1_suffix.png
| - |- image2_suffix.png
| - |- image3_suffix.png
**At this point the images are not aligned, but training can be done using the current method**
# Scenes
1. When you need to explicitly select a part of the data from the sample for training, you can only remove those bad labels without changing the image.
2. When there are some problems in the data set, such as some dirty data, the wrong label can be deleted without considering the alignment problem as before
3. Finding images through labels is more in line with common sense, but finding labels through images is obviously not, because the existence of unlabeled images in the data set itself should be reasonable
# note
This PR was indeed used by me during training, my usage scenario:
I use a large data set, about 12w of data, I don't want to map all the labels to grayscale images to initiate training, so I modified the code.

Everyone's needs may be different, but I think this modification itself will not hurt any other parts, so I made this PR

mmseg/datasets/custom.py
```diff
@@ -164,13 +164,14 @@ def load_annotations(self, img_dir, img_suffix, ann_dir, seg_map_suffix,
                 img_infos.append(img_info)
         else:
             for img in self.file_client.list_dir_or_file(
```
MeowZheng, line 166:
```suggestion
            for seg_map in self.file_client.list_dir_or_file(
```

mmseg/datasets/custom.py
```diff
@@ -164,13 +164,14 @@ def load_annotations(self, img_dir, img_suffix, ann_dir, seg_map_suffix,
...
-                    dir_path=img_dir,
+                    dir_path=ann_dir,
```
MeowZheng, line 167: how do we handle `ann_dir=None`?

Bin-ze: if need handle ann_dir = None
can：
assert ann_dir==None: 'please add anntations file!'
Is this possible?

Bin-ze: This PR is just some suggestions from me, if you think it is not necessary at all, you can just close it

MeowZheng: I think it would be better to still get image path first, and get seg map path by replace the suffix and check whether seg map exists by

```python
if self.file_client(seg_map):
    img_info['ann'] = dict(seg_map=seg_map)
else:
    continue
```
what do you think about this modification?

PR #2402, opened 2022-12-12
[CI] Add torch1.13 checking in CI

xiexinch:
## Motivation

Ref: https://github.com/open-mmlab/mmcv/pull/2441

Add torch1.13 checking in CI.

PR #2405, opened 2022-12-12
PR for fix ci

ice-tong:


ice-tong:
We found the solution: https://github.com/open-mmlab/mmsegmentation/pull/2402#issuecomment-1346863784

Issue #1558, opened 2022-05-05
Custom datasets 


caihaunqai:
Hello, I have followed the Tutorial 2 to customize my own dataset. It can find all the pictures and train normally. However, the accuracy is 99.99% at the first iter, since it estimate all pixels to one class (two class in total). I use some saved model to test and find that it really estimate all pixels to one class. I checked my dataset and the labels for GT are 0 or 1， which are correct. But I still can not get the true results. Do you know why? Thank you~

caihaunqai:
I print one dataset and it shows like the following picture. I don't know why it has 255 since I check my dataset and label for GT are 0 or 1.
![image](https://user-images.githubusercontent.com/56779804/167058792-8de19f12-fe8e-4cf9-a018-92d1ab235dea.png)

MeowZheng:
Might check `ignore_index` in decode_head cfg, as if `ignore_index=0` will ingore all 0 label

caihaunqai:
Thank you for your reply. I have check the dataset file in "mmseg/datasets" and found ignore_index is 255 in default. Is it the&nbsp;variable that you mentioned?&nbsp;
&nbsp;


------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "open-mmlab/mmsegmentation"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年5月6日(星期五) 下午4:51
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [open-mmlab/mmsegmentation] Custom datasets  (Issue #1558)





 
Might check ignore_index in decode_head cfg, as if ignore_index=0 will ingore all 0 label

MeowZheng:
> I print one dataset and it shows like the following picture. I don't know why it has 255 since I check my dataset and label for GT are 0 or 1. ![image](https://user-images.githubusercontent.com/56779804/167058792-8de19f12-fe8e-4cf9-a018-92d1ab235dea.png)

I found the label of dataset is 255, it might be the problem of dataset pre-processes. You might try to covert the color label to seg map label, ref https://github.com/open-mmlab/mmsegmentation/blob/ee67c808b2ac9dad38e082d168741dd1097ffc93/tools/convert_datasets/isaid.py#L14-L45

caihaunqai:
[...]

PR #2437, opened 2022-12-27
CodeCamp #1558 [Doc] Add pspnet model structure graph

haofengsiji:
## Motivation

Add pspnet model structure based on the 超级视客营 activity.

## Modification

Add pspnet model structure graph in pspnet config readme.


RangeKing (approved):
LGTM

PR #2500, opened 2023-01-19
[Fix] Switch order of `reduce_zero_label` and applying `label_map`

siddancha:
## Motivation

I want to fix a bug through this PR. The bug occurs when two options -- `reduce_zero_label=True`, and custom classes are used. `reduce_zero_label` remaps the GT seg labels by remapping the zero-class to 255 which is ignored. Conceptually, this should occur *before* the `label_map` is applied, which maps *already reduced labels*. However, currently, the `label_map` is applied before the zero label is reduced.

## Modification

The modification is simple:
- I've just interchanged the order of the two operations by moving 4 lines from bottom to top.
- I've added a test that passes when the fix is introduced, and fails on the original `master` branch.

## BC-breaking (Optional)

I do not anticipate this change braking any backward-compatibility.

## Checklist

- [x] Pre-commit or other linting tools are used to fix the potential lint issues.
  - _I've fixed all linting/pre-commit errors._
- [x] The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
  - _I've added a unit test._ 
- [x] If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMDet3D.
  - _I don't think this change affects MMDet or MMDet3D._
- [x] The documentation has been modified accordingly, like docstring or example tutorials.
  - _This change fixes an existing bug and doesn't require modifying any documentation/docstring._

Issue #1415, opened 2022-03-24
`label_map` used twice during evaluation


panfeng-hover:
I write a customized dataset, where the label ids are `(0, 1, 2, 3, 4, 5, 6)`, and I would like to merge some classes, so I set 

1) `self.custom_classes = True`
2) `self.label_map ={4: 3, 6: 3, 5: 4}`. 
Namely convert label ids of 4 & 6 in the ground truth to 3, and convert label id 5 to 4, as mmsegmentation is assuming the label ids is [continuous](https://github.com/open-mmlab/mmsegmentation/blob/01e35aa6f77694e5fbabf4b2593e11130ddc1898/mmseg/core/evaluation/metrics.py#L79),

However, it looks like the `label_map` is wrongly used during evaluation, in details,

the [`self.get_gt_seg_map_by_idx(index)`](https://github.com/open-mmlab/mmsegmentation/blob/01e35aa6f77694e5fbabf4b2593e11130ddc1898/mmseg/datasets/custom.py#L287) used here already converted the ground truth label ids from `(0, 1, 2, 3, 4, 5, 6)` to `(0, 1, 2, 3, 4)`, but during later [`intersect_and_union`](https://github.com/open-mmlab/mmsegmentation/blob/01e35aa6f77694e5fbabf4b2593e11130ddc1898/mmseg/core/evaluation/metrics.py#L68), the label_map is used again, so the ground truth label ids will become `(0, 1, 2, 3)` from `(0, 1, 2, 3, 4)`.

panfeng-hover:
@MengzhangLI would u like to take a look? Thanks!

panfeng-hover:
It is similar to the one used in [`evaluate`](https://github.com/open-mmlab/mmsegmentation/blob/01e35aa6f77694e5fbabf4b2593e11130ddc1898/mmseg/datasets/custom.py#L408) function, where `gt_seg_maps = self.get_gt_seg_maps()` is already converted by the `label_map` during [`pre_pipeline`](https://github.com/open-mmlab/mmsegmentation/blob/01e35aa6f77694e5fbabf4b2593e11130ddc1898/mmseg/datasets/custom.py#L261)

PR #2504, opened 2023-01-20
[Fix] Fix `reduce_zero_label` in evaluation

siddancha:
## Motivation

Through this PR, I (1) fix a bug, and (2) perform some associated cleanup, and (3) add a unit test. The bug occurs during evaluation when two options -- `reduce_zero_label=True`, and custom classes are used. The bug was that the `reduce_zero_label` is not properly propagated (see details below).

## Modification

1. **Bugfix**

The bug occurs [in the initialization of `CustomDataset`](https://github.com/open-mmlab/mmsegmentation/blob/5d49918b3c48df5544213562aa322bfa89d67ef1/mmseg/datasets/custom.py#L108-L110) where the `reduce_zero_label` flag is not propagated to its member `self.gt_seg_map_loader_cfg`:

```python
self.gt_seg_map_loader = LoadAnnotations(
) if gt_seg_map_loader_cfg is None else LoadAnnotations(
    **gt_seg_map_loader_cfg)
```

Because the `reduce_zero_label` flag was not being propagated, the zero label reduction was being [unnecessarily and explicitly duplicated during the evaluation](https://github.com/open-mmlab/mmsegmentation/blob/5d49918b3c48df5544213562aa322bfa89d67ef1/mmseg/core/evaluation/metrics.py#L66-L69).

As pointed in a previous PR (#2500), `reduce_zero_label` must occur before applying the `label_map`. Due to this bug, the order gets reversed when both features are used simultaneously.

This has been fixed to:

```python
self.gt_seg_map_loader = LoadAnnotations(
    reduce_zero_label=reduce_zero_label, **gt_seg_map_loader_cfg)
```

2. **Cleanup**

Due to the bug fix, since both `reduce_zero_label` and `label_map` are being applied in `get_gt_seg_map_by_idx()` (i.e. `LoadAnnotations.__call__()`), the evaluation does not need perform them anymore.

This was pointed out for `label_map` in a previous issue (#1415) that the `label_map` should  not be applied in the evaluation. This was handled by [passing an empty dict](https://github.com/open-mmlab/mmsegmentation/blob/5d49918b3c48df5544213562aa322bfa89d67ef1/mmseg/datasets/custom.py#L306-L311):

```python
# as the labels has been converted when dataset initialized
# in `get_palette_for_custom_classes ` this `label_map`
# should be `dict()`, see
# https://github.com/open-mmlab/mmsegmentation/issues/1415
# for more ditails
label_map=dict(),
reduce_zero_label=self.reduce_zero_label))
```

Similar to this, I now also set `reduce_label=False` since it is now also being handled by `get_gt_seg_map_by_idx()` (i.e. `LoadAnnotations.__call__()`).

3. **Unit test**

I've added a unit test that tests the `CustomDataset.pre_eval()` function when `reduce_zero_label=True` and custom classes are used. The test fails on the original `master` branch but passes with this fix.

## BC-breaking (Optional)

I do not anticipate this change braking any backward-compatibility.

## Checklist

  - _I've added a test that passes when the fix is introduced, and fails on the original master branch._

### DIFF
--- a/mmseg/datasets/custom.py
+++ b/mmseg/datasets/custom.py
@@ -301,14 +301,7 @@ class CustomDataset(Dataset):
                     pred,
                     seg_map,
                     len(self.CLASSES),
-                    self.ignore_index,
-                    # as the labels has been converted when dataset initialized
-                    # in `get_palette_for_custom_classes ` this `label_map`
-                    # should be `dict()`, see
-                    # https://github.com/open-mmlab/mmsegmentation/issues/1415
-                    # for more ditails
-                    label_map=dict(),
-                    reduce_zero_label=False))
+                    self.ignore_index))
 
         return pre_eval_results
 
@@ -424,9 +417,7 @@ class CustomDataset(Dataset):
                 gt_seg_maps,
                 num_classes,
                 self.ignore_index,
-                metric,
-                label_map=dict(),
-                reduce_zero_label=self.reduce_zero_label)
+                metric)
         # test a list of pre_eval_results
         else:
             ret_metrics = pre_eval_to_metrics(results, metric)
--- a/mmseg/datasets/custom.py
+++ b/mmseg/datasets/custom.py
@@ -297,11 +297,8 @@ class CustomDataset(Dataset):
         for pred, index in zip(preds, indices):
             seg_map = self.get_gt_seg_map_by_idx(index)
             pre_eval_results.append(
-                intersect_and_union(
-                    pred,
-                    seg_map,
-                    len(self.CLASSES),
-                    self.ignore_index))
+                intersect_and_union(pred, seg_map, len(self.CLASSES),
+                                    self.ignore_index))
 
         return pre_eval_results
 
@@ -412,12 +409,8 @@ class CustomDataset(Dataset):
             if gt_seg_maps is None:
                 gt_seg_maps = self.get_gt_seg_maps()
             num_classes = len(self.CLASSES)
-            ret_metrics = eval_metrics(
-                results,
-                gt_seg_maps,
-                num_classes,
-                self.ignore_index,
-                metric)
+            ret_metrics = eval_metrics(results, gt_seg_maps, num_classes,
+                                       self.ignore_index, metric)
         # test a list of pre_eval_results
         else:
             ret_metrics = pre_eval_to_metrics(results, metric)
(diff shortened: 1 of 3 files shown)


siddancha:
This PR cleans up some unnecessary arguments in evaluation functions. This needs to update the documentation/docstring that I  don't have access to.

MeowZheng:
Thanks for your contribution. @MengzhangLI please have a look

MeowZheng (commented):
My consideration is that the modifications of `intersect_and_union` and other evaluation functions are bc-breaking for the users who use it alone, but I support the modification of `CustomDataset`. I suggest keeping input arguments of evaluation functions.

siddancha:
@MeowZheng sounds good! I've reverted my "cleanup" in `metrics.py` for backwards compatibility in 
ca3a7b2 . Now, I keep the `label_dict` and `return_label` arguments in the evaluation functions, but when calling them from `CustomDataset`, I set `return_label=False` just like setting `label_map=dict()`.

mmseg/datasets/custom.py
```diff
@@ -66,8 +66,8 @@ class CustomDataset(Dataset):
             The palette of segmentation map. If None is given, and
             self.PALETTE is None, random palette will be generated.
             Default: None
-        gt_seg_map_loader_cfg (dict, optional): build LoadAnnotations to
-            load gt for evaluation, load from disk by default. Default: None.
+        gt_seg_map_loader_cfg (dict): build LoadAnnotations to load gt for
+            evaluation, load from disk by default. Default: empty.
```
MeowZheng, line 70:
```suggestion
            evaluation, load from disk by default. Default: ``dict()``.
```

MeowZheng (commented):
Lint failed, as there is some problem from isort https://github.com/PyCQA/isort/issues/2077, but it has been fixed in https://github.com/PyCQA/isort/pull/2078.

We are working on upgrading the isort, but meet other problems now. :(

MeowZheng:
We have fixed lint problem of master branch #2520
I suggest rebasing your modification on the latest code of master

siddancha:
@MeowZheng I rebased this PR on the latest version of `master`.

MeowZheng (approved):

MengzhangLI (approved):

MengzhangLI:
Hi, @siddancha many thanks for your nice PR and detailed description!

PR #2516, opened 2023-01-28
[Fix] Unfinished label conversion from `-1` to `255` in 1.x

siddancha:
## Motivation

This is motivated by a previously unfinished PR (#2332). In that PR, the label -1 was changed to 255 in `BaseSegDataset`, which is correct. However, it was changed at only one location. There is another location in `mmseg/datasets/basesegdataset.py` where -1 was still being used that was not converted to 255. I have now converted it to 255.

This is exactly same as a similar fix to the `master` branch via #2515 .

## Modification

I've simply converted the snipped

```python
if new_id != -1:
    new_palette.append(palette[old_id])
```
to 
```python
if new_id != 255:
    new_palette.append(palette[old_id])
```

## Checklist

  - _No unit tests need to be added or were affected.

### DIFF
--- a/mmseg/datasets/basesegdataset.py
+++ b/mmseg/datasets/basesegdataset.py
@@ -220,7 +220,7 @@ class BaseSegDataset(BaseDataset):
             # return subset of palette
             for old_id, new_id in sorted(
                     self.label_map.items(), key=lambda x: x[1]):
-                if new_id != -1:
+                if new_id != 255:
                     new_palette.append(palette[old_id])
             new_palette = type(palette)(new_palette)
         else:


MeowZheng (approved):

MeowZheng:
This modification looks good to me, and the ci failed probably because upstream repos mmdet or mmengine did some backward incompatible changes, and we are working on fix it.

MeowZheng:
we have fixed ci problem in #2519  and please rebase the modification on it. :)

MengzhangLI (approved):

siddancha:
Great! Merged the latest version of `dev-1.x`.

PR #2517, opened 2023-01-29
[Fix] Switch order of `reduce_zero_label` and applying `label_map` in 1.x

siddancha:
This is an almost exact duplicate of #2500 (that was made to the `master` branch) now applied to the `1.x` branch.

---

## Motivation

I want to fix a bug through this PR. The bug occurs when two options -- `reduce_zero_label=True`, and custom classes are used. `reduce_zero_label` remaps the GT seg labels by remapping the zero-class to 255 which is ignored. Conceptually, this should occur *before* the `label_map` is applied, which maps *already reduced labels*. However, currently, the `label_map` is applied before the zero label is reduced.

## Modification

The modification is simple:
- I've just interchanged the order of the two operations by moving a few lines from bottom to top.
- I've added a test that passes when the fix is introduced, and fails on the original `master` branch.

## BC-breaking (Optional)

I do not anticipate this change braking any backward-compatibility.

## Checklist

  - _I've added a unit test._ 

### DIFF
--- a/mmseg/datasets/transforms/loading.py
+++ b/mmseg/datasets/transforms/loading.py
@@ -96,14 +96,6 @@ class LoadAnnotations(MMCV_LoadAnnotations):
             img_bytes, flag='unchanged',
             backend=self.imdecode_backend).squeeze().astype(np.uint8)
 
-        # modify if custom classes
-        if results.get('label_map', None) is not None:
-            # Add deep copy to solve bug of repeatedly
-            # replace `gt_semantic_seg`, which is reported in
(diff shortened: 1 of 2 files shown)


MeowZheng (approved):
we are still working on fixing ci. :(

MengzhangLI (approved):

MeowZheng:
we have fixd ci problem in #2519 and please rebase this modification on it

siddancha:
Great! Merged the latest version of `dev-1.x`.

PR #2535, opened 2023-01-30
[Types] Fix misleading `List[Tensor]` types

siddancha:
## Motivation

A few functions in the following files:
- `decode_head.py`
- `encoder_decoder.py`
- `cascade_encoder_decoder.py`
 
have a return type annotation as `List[Tensor]` whereas their actual return type is `Tensor`. This can be seen by following the trace of what each functions return internally, which eventually leads to `seg_logits: Tensor`.

This type annotation is misleading and confusing when a user is trying to read and understand the code.

## Modification

1. Fix type annotations from `List[Tensor]` to `Tensor`.
2. Make a return type more specific from `list` to `SampleList`.

## Checklist

  - _This involves only changing type annotations, so there is no need for tests._ 
- [ ] The documentation has been modified accordingly, like docstring or example tutorials.
  - _This change fixes existing type annotations, so it may require a change in the documentation/docstring that I don't have access to._

### DIFF
--- a/mmseg/models/decode_heads/decode_head.py
+++ b/mmseg/models/decode_heads/decode_head.py
@@ -263,7 +263,7 @@ class BaseDecodeHead(BaseModule, metaclass=ABCMeta):
         return losses
 
     def predict(self, inputs: Tuple[Tensor], batch_img_metas: List[dict],
-                test_cfg: ConfigType) -> List[Tensor]:
+                test_cfg: ConfigType) -> Tensor:
         """Forward function for prediction.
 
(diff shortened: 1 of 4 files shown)


MeowZheng (approved):

MengzhangLI (approved):

siddancha:
Looks like I accidentally created this PR for `1.x` branch, and not `dev-1.x`. Is this a problem? Should we merge `1.x` to `dev-`1.x` so that both are synced?

PR #2653, opened 2023-02-27
[Fix] Fix pretrained models of SegNeXt in master branch.

MengzhangLI:
## Motivation

Transfer keys of each `mscan_x.pth` pretrained models of SegNeXt, and upload them in the website.

The reason of transferring keys is we modify original repo [`.dwconv.dwconv.xxx`](https://github.com/Visual-Attention-Network/SegNeXt/blob/main/mmseg/models/backbones/mscan.py#L21) to [`.dwconv.xxx`](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/models/backbones/mscan.py#L43).

### DIFF
--- a/configs/segnext/README.md
+++ b/configs/segnext/README.md
@@ -31,10 +31,6 @@ We present SegNeXt, a simple convolutional network architecture for semantic seg
 }
 ```
 
-## Pretrained model
-
-The pretrained model could be found [here](https://cloud.tsinghua.edu.cn/d/c15b25a6745946618462/) from [original repo](https://github.com/Visual-Attention-Network/SegNeXt). You can download and put them in `./pretrain` folder.
-
 ## Results and models
 
 ### ADE20K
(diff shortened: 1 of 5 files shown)


configs/segnext/README.md
```diff
@@ -48,6 +44,8 @@ The pretrained model could be found [here](https://cloud.tsinghua.edu.cn/d/c15b2
 
 Note:
 
+- The keys of our provided pretrained models are different from [original weights](https://cloud.tsinghua.edu.cn/d/c15b25a6745946618462/) released in [original repo](https://github.com/Visual-Attention-Network/SegNeXt), please use them correctly.
```
MeowZheng, line 47: When we integrated SegNeXt into MMSegmentation, we modified some layers' names to make them more precise and concise without changing the model architecture. Therefore, the keys of pre-trained weights are different from [the original weights](https://cloud.tsinghua.edu.cn/d/c15b25a6745946618462/), but don't worry about these changes. we have converted them and uploaded the checkpoints, you might fine URL of pre-trained checkpoints in config files and can use them directly for training.

MeowZheng (approved):

PR #2738, opened 2023-03-10
Add support for sample weighting (dataset imbalance)

InakiRaba91:

## Motivation

Sample weighting is a common strategy used to deal with imbalanced datasets. The idea is to weight the contribution to the loss of each item on the dataset based to its frequency.

There are different strategies, such as Inverso of Number of Samples (INS), Inverse of Square Root of Number of Samples (ISNS) or Effective Number of Samples ([ENS](https://openaccess.thecvf.com/content_CVPR_2019/papers/Cui_Class-Balanced_Loss_Based_on_Effective_Number_of_Samples_CVPR_2019_paper.pdf)), to cite a few (see a brief summary [here](https://medium.com/gumgum-tech/handling-class-imbalance-by-introducing-sample-weighting-in-the-loss-function-3bdebd8203b4))

## Modification

The code already supports providing an item-wise weight to the loss. However, that weight is not exposed at the level of the head, so it can not be injected at runtime.

There are mechanisms to weight different classes, or different losses (multi-loss scenario). But as far as I understand, it is not possible to carry out sample weighting.

## BC-breaking (Optional)

No, the changes only extend the functionality without affecting any current scenario.

## Use cases (Optional)

Support added for sample weighting. In order to do so, the dataset just needs to add a float scalar named "weight" to the annotation info.

The loading pipeline has been uploaded to pick it up, and it will be injected to the loss function.

This should be quite useful when dealing with imbalanced datasets.

## Checklist

## Issue

https://github.com/open-mmlab/mmdetection/issues/9905

### DIFF
--- a/mmseg/datasets/pipelines/loading.py
+++ b/mmseg/datasets/pipelines/loading.py
@@ -147,6 +147,8 @@ class LoadAnnotations(object):
             gt_semantic_seg_copy = gt_semantic_seg.copy()
             for old_id, new_id in results['label_map'].items():
                 gt_semantic_seg[gt_semantic_seg_copy == old_id] = new_id
+        if "weight" in results["ann_info"]:
+            results["weight"] = results["ann_info"]["weight"] 
         results['gt_semantic_seg'] = gt_semantic_seg
         results['seg_fields'].append('gt_semantic_seg')
         return results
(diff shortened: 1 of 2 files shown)


InakiRaba91:
Please let me know if you can think of any issues with this approach. If you believe it is on the right track, I'll update the forward_train method in point_head.py and add unit tests for the new functionality.

MeowZheng:
thanks for your contribution, we are working on reviewing this pr.

PR #2783, opened 2023-03-22
Add support for the focal Tversky loss

zifuwanggg:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

The focal Tversky loss was proposed in https://arxiv.org/abs/1810.07842. It has nearly 600 citations and has been shown to be extremely useful for highly imbalanced (medical) datasets. To add support for the focal Tversky loss, only few lines of changes are needed for the Tversky loss.

## Modification

Add `gamma` as (optional) argument in the constructor of `TverskyLoss`. This parameter is then passed to `tversky_loss` to compute the focal Tversky loss.

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repos?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects, like MMDet or MMDet3D.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #2791, opened 2023-03-24
[Feature] Add support for the focal Tversky loss

zifuwanggg:

## Motivation

The focal Tversky loss was proposed in https://arxiv.org/abs/1810.07842. It has nearly 600 citations and has been shown to be extremely useful for highly imbalanced (medical) datasets. To add support for the focal Tversky loss, only few lines of changes are needed for the Tversky loss.

## Modification

Add `gamma` as (optional) argument in the constructor of `TverskyLoss`. This parameter is then passed to `tversky_loss` to compute the focal Tversky loss. 

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

Reopening of previous [PR](https://github.com/open-mmlab/mmsegmentation/pull/2783).

### DIFF
--- a/mmseg/models/losses/tversky_loss.py
+++ b/mmseg/models/losses/tversky_loss.py
@@ -31,7 +31,6 @@ def tversky_loss(pred,
                 valid_mask=valid_mask,
                 alpha=alpha,
                 beta=beta,
-                gamma=gamma,
                 smooth=smooth)
             if gamma > 1.0:
                 tversky_loss **= gamma
--- a/mmseg/models/losses/tversky_loss.py
+++ b/mmseg/models/losses/tversky_loss.py
@@ -68,7 +68,7 @@ class TverskyLoss(nn.Module):
     <https://arxiv.org/abs/1706.05721>`
     and `A novel focal Tversky loss function with improved attention U-Net for
     lesion segmentation.
-    
+
     <https://arxiv.org/abs/1810.07842>`_.
     Args:
         smooth (float): A float number to smooth loss, and avoid NaN error.
--- a/mmseg/models/losses/tversky_loss.py
+++ b/mmseg/models/losses/tversky_loss.py
@@ -16,6 +16,7 @@ def tversky_loss(pred,
                  valid_mask,
                  alpha=0.3,
                  beta=0.7,
+                 gamma=1.0,
                  smooth=1,
                  class_weight=None,
                  ignore_index=255):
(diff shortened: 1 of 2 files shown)


mmseg/models/losses/tversky_loss.py
```diff
@@ -87,15 +93,18 @@ def __init__(self,
...
                  beta=0.7,
+                 gamma=1.0,
                  loss_name='loss_tversky'):
         super(TverskyLoss, self).__init__()
         self.smooth = smooth
         self.class_weight = get_class_weight(class_weight)
         self.loss_weight = loss_weight
         self.ignore_index = ignore_index
         assert (alpha + beta == 1.0), 'Sum of alpha and beta but be 1.0!'
+        assert gamma >= 1.0, 'gamma should be at least 1.0!'
```
xiexinch, line 104: Since there is an assertion statement, we might add a unit test for it.

mmseg/models/losses/tversky_loss.py
```diff
@@ -31,6 +32,8 @@ def tversky_loss(pred,
                 alpha=alpha,
                 beta=beta,
                 smooth=smooth)
+            if gamma > 1.0:
+                tversky_loss **= gamma
```
xiexinch, line 36: Should it be `1 / gamma`?
At the official implementation https://github.com/nabsabraham/focal-tversky-unet/blob/master/losses.py#L67, they calculate the FTL with `gamma=1/(4/3)=0.75`

xiexinch (commented):
Hi @zifuwanggg,
Thanks for your contribution! We really appreciate it. Here are some comments that should be resolved, please take a view.

mmseg/models/losses/tversky_loss.py
zifuwanggg: Yes indeed. I've modified it to `1 / gamma`.

mmseg/models/losses/tversky_loss.py
```diff
@@ -75,6 +78,9 @@ class TverskyLoss(nn.Module):
         beta (float, in [0, 1]):
             The coefficient of false negatives. Default: 0.7.
             Note: alpha + beta = 1.
+        gamma (float, in [1, inf]): The focal term. When `gamma` > 1,
+            the loss focuses more on less accurate predictions that
+            have been misclassified. Default: 1.0.
```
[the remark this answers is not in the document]
zifuwanggg: I've added the reference in the docstring.

mmseg/models/losses/tversky_loss.py
zifuwanggg: I presume I don't need to do anything for this comment?

PR #2966, opened 2023-04-29
Prevent passed `ann_file` from silently failing to load

haruishi43:

## Motivation

While customizing the number of samples using `ann_file` for Cityscapes, I noticed that when the `ann_file` name is incorrect, it will silently resort to loading the dataset from the directory.
I think when the user intends to load using `ann_file`, it should not silently fail, but give some sort of error message or warning.

## Modification

I added assertion to check whether the `ann_file` exists instead of silently resorting to loading from the directory.
Since `ann_file` is set to `''` by default and joined with `self.data_root`, I used `osp.isdir` to first check if `self.ann_dir` is a directory or text file.

## BC-breaking (Optional)

Not that I am aware of.

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmseg/datasets/basesegdataset.py
+++ b/mmseg/datasets/basesegdataset.py
@@ -235,7 +235,10 @@ class BaseSegDataset(BaseDataset):
         data_list = []
         img_dir = self.data_prefix.get('img_path', None)
         ann_dir = self.data_prefix.get('seg_map_path', None)
-        if osp.isfile(self.ann_file):
+        if not osp.isdir(self.ann_file):
+            assert osp.isfile(self.ann_file), (
+                f'Failed to load `ann_file` {self.ann_file}'
+            )
             lines = mmengine.list_from_file(
                 self.ann_file, backend_args=self.backend_args)
             for line in lines:
--- a/mmseg/datasets/basesegdataset.py
+++ b/mmseg/datasets/basesegdataset.py
@@ -235,9 +235,9 @@ class BaseSegDataset(BaseDataset):
         data_list = []
         img_dir = self.data_prefix.get('img_path', None)
         ann_dir = self.data_prefix.get('seg_map_path', None)
-        if not osp.isdir(self.ann_file):
-            assert osp.isfile(
-                self.ann_file), (f'Failed to load `ann_file` {self.ann_file}')
+        if not osp.isdir(self.ann_file) and self.ann_file:
+            assert osp.isfile(self.ann_file), \
+                f'Failed to load `ann_file` {self.ann_file}'
             lines = mmengine.list_from_file(
                 self.ann_file, backend_args=self.backend_args)
             for line in lines:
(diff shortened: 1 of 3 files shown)


mmseg/datasets/basesegdataset.py
```diff
@@ -235,7 +235,9 @@ def load_data_list(self) -> List[dict]:
         data_list = []
         img_dir = self.data_prefix.get('img_path', None)
         ann_dir = self.data_prefix.get('seg_map_path', None)
-        if osp.isfile(self.ann_file):
+        if not osp.isdir(self.ann_file) and self.ann_file:
```
csatsurnh, line 238: seems not right when `ann_file` is a `dir`

haruishi43: If I'm understanding correctly, `ann_file` is not a directory, but some `txt` file for mmsegmentation.
If we want to load from some directory, we would use `ann_dir` and `img_dir` right?

csatsurnh: yes, by default `ann_file` should be a file or '', but I think we need to take `dir` into consideration, just in case.

haruishi43: The old code also ignores the case when `ann_file` is a directory.
I'm not quite sure what you want to do here... Since `ann_file` already has the name `file`, it is assumed that the it is a annotation file rather than a directory. When `ann_file` is `''`, it will automatically turn into a directory anyway since mmengine joins `ann_file` with `data_root`.

csatsurnh: actually the origin version would lead to `else` if `ann_file` is not a file, what I would like to recommend to do was just to make sure all the cases would work correctly since we can not guarantee all the cases are used fully on the default manner.

haruishi43: Yes, that is the same as in the revised code. If `ann_file` is not a file, it would lead to `else` as well. Except when `ann_file` is a file, it will raise an error if the file exists or not instead of silently failing and going to `else`.
I had a problem where I have multiple servers and sometimes I forget to make a specific `ann_file` and it will silently load all of the directory instead. I think failing is much better than silently going to `else` in the case where `ann_file` is given.

csatsurnh: it seems that it would go on top but not else if it's a dir?

csatsurnh: yes I understand your point, just need to get it right when it's a dir

haruishi43: The `not osp.isdir(self.ann_file)` should check this and would go to the `else`, right?

csatsurnh: oops, so sorry, I made a mistake

csatsurnh (approved):

mmseg/datasets/basesegdataset.py
haruishi43: @csatsurnh  No problem! Thanks for reviewing this.

PR #3044, opened 2023-05-25
[Enhance] Enable full precision training on Ascend NPU.

Ginray:
## Motivation

We will support full precision training on the next generation Ascend NPU, so there is no need to enable mixed precision by default.

## Modification

Determine whether the current chip supports full precision training, and automatically enable mixed precision.

## BC-breaking (Optional)

Not affected.

## Use cases (Optional)

We have verified the correctness on the Ascend NPU.


xiexinch:
Hi @Ginray,
I'm sorry that we won't maintain the master branch anymore.

PR #3152, opened 2023-06-30
Update seg tta py

ZiAn-Su:
## Motivation

When using the - tta command for multi-scale prediction, and the test set is not annotated, although format_only has been set true in test_evaluator,  but SegTTAModel class  still threw error 'AttributeError: 'SegDataSample' object has no attribute '_gt_sem_seg''. 

## Modification

The reason is SegTTAModel  didn't determine if there were annotations in the dataset, so I added the code to make the judgment and let the program run normally on my computer.

### DIFF
--- a/mmseg/models/segmentors/seg_tta.py
+++ b/mmseg/models/segmentors/seg_tta.py
@@ -39,11 +39,18 @@ class SegTTAModel(BaseTTAModel):
                             ).to(logits).squeeze(1)
             else:
                 seg_pred = logits.argmax(dim=0)
-            data_sample = SegDataSample(
-                **{
-                    'pred_sem_seg': PixelData(data=seg_pred),
-                    'gt_sem_seg': data_samples[0].gt_sem_seg
-                })
+            if hasattr(data_samples[0], 'gt_sem_seg'): 
+                data_sample = SegDataSample(
+                    **{
+                        'pred_sem_seg': PixelData(data=seg_pred),
+                        'gt_sem_seg': data_samples[0].gt_sem_seg
+                    })
+            else:
+                data_sample = SegDataSample(
+                    **{
+                        'pred_sem_seg': PixelData(data=seg_pred),
+                        'gt_sem_seg': PixelData(data=seg_pred)
+                    }) 
             data_sample.set_metainfo({'img_path': data_samples[0].img_path})
             predictions.append(data_sample)
         return predictions
--- a/mmseg/models/segmentors/seg_tta.py
+++ b/mmseg/models/segmentors/seg_tta.py
@@ -39,7 +39,7 @@ class SegTTAModel(BaseTTAModel):
                             ).to(logits).squeeze(1)
             else:
                 seg_pred = logits.argmax(dim=0)
-            if hasattr(data_samples[0], 'gt_sem_seg'): 
+            if hasattr(data_samples[0], 'gt_sem_seg'):
                 data_sample = SegDataSample(
                     **{
                         'pred_sem_seg': PixelData(data=seg_pred),
@@ -50,7 +50,7 @@ class SegTTAModel(BaseTTAModel):
                     **{
                         'pred_sem_seg': PixelData(data=seg_pred),
                         'gt_sem_seg': PixelData(data=seg_pred)
-                    }) 
+                    })
             data_sample.set_metainfo({'img_path': data_samples[0].img_path})
             predictions.append(data_sample)
         return predictions


mmseg/models/segmentors/seg_tta.py
```diff
@@ -39,11 +39,18 @@ def merge_preds(self, data_samples_list: List[SampleList]) -> SampleList:
...
+                data_sample = SegDataSample(
+                    **{
+                        'pred_sem_seg': PixelData(data=seg_pred),
+                        'gt_sem_seg': data_samples[0].gt_sem_seg
+                    })
+            else:
+                data_sample = SegDataSample(
+                    **{
+                        'pred_sem_seg': PixelData(data=seg_pred),
+                        'gt_sem_seg': PixelData(data=seg_pred)
```
xiexinch, line 52: Can we delete this field?

xiexinch (approved):

mmseg/models/segmentors/seg_tta.py
ZiAn-Su: I'm afraid not. seg_pred is the final prediction, we will get the last prediction in data_samples if we delete it. But we can still simplify the code by replacing
```suggestion
                       if hasattr(data_samples[0], 'gt_sem_seg'): 
                data_sample = SegDataSample(
                    **{
                        'pred_sem_seg': PixelData(data=seg_pred),
                        'gt_sem_seg': data_samples[0].gt_sem_seg
                    })
            else:
                data_sample = SegDataSample(
                    **{
                        'pred_sem_seg': PixelData(data=seg_pred),
                        'gt_sem_seg': PixelData(data=seg_pred)
                    })
```
 with
```suggestion
            data_sample.set_data({'pred_sem_seg': PixelData(data=seg_pred)})
            if hasattr(data_samples[0], 'gt_sem_seg'):
                data_sample.set_data({'gt_sem_seg': data_samples[0].gt_sem_seg})
```

mmseg/models/segmentors/seg_tta.py
```diff
@@ -39,11 +39,18 @@ def merge_preds(self, data_samples_list: List[SampleList]) -> SampleList:
                             ).to(logits).squeeze(1)
             else:
                 seg_pred = logits.argmax(dim=0)
-            data_sample = SegDataSample(
-                **{
-                    'pred_sem_seg': PixelData(data=seg_pred),
-                    'gt_sem_seg': data_samples[0].gt_sem_seg
-                })
```
ZiAn-Su, line 46: I'm afraid not. seg_pred is the final prediction, we will get the last prediction in data_samples if we delete it. But we can still simplify the code by
```
data_sample.set_data({'pred_sem_seg': PixelData(data=seg_pred)})
if hasattr(data_samples[0], 'gt_sem_seg'):
    data_sample.set_data({'gt_sem_seg': data_samples[0].gt_sem_seg})
```

mmseg/models/segmentors/seg_tta.py
ZiAn-Su: I'm afraid not. seg_pred is the final prediction, we will get the last prediction in data_samples if we delete it. But we can still simplify the code by replacing
```
if hasattr(data_samples[0], 'gt_sem_seg'):
    data_sample = SegDataSample(
        **{
            'pred_sem_seg': PixelData(data=seg_pred),
            'gt_sem_seg': data_samples[0].gt_sem_seg
        })
else:
    data_sample = SegDataSample(
        **{
            'pred_sem_seg': PixelData(data=seg_pred),
            'gt_sem_seg': PixelData(data=seg_pred)
```
with
```
data_sample.set_data({'pred_sem_seg': PixelData(data=seg_pred)})
if hasattr(data_samples[0], 'gt_sem_seg'):
    data_sample.set_data({'gt_sem_seg': data_samples[0].gt_sem_seg})
```

xiexinch: Sorry I wasn't clear, I just wanted to delete the field `gt_sem_seg`, because we want to make sure the semantics of these variables are correct.

ZiAn-Su: Sorry to misunderstood. Sure, very rigorous idea, we should do this.  And the less equivalent code has also been tested, it may be more concise.

shallweiwei:
I also encountered this bug

shallweiwei:
I would add that I can still see the progress bar.It is because there are too many warnings……

ZiAn-Su:
> I also encountered this bug.I changed the code with this method, but I can't see the progress bar now. And the output is
> 
> ```python
> envs/openmmlab/lib/python3.8/site-packages/mmengine/structures/pixel_data.py:83: UserWarning: The shape of value will convert from torch.Size([998, 998]) to torch.Size([1, 998, 998])
> ```
I haven't encountered this problem, perhaps you can try increasing the zoom ratio in tta_pipeline to avoid the image being too small.

ZiAn-Su:
> I would add that I can still see the progress bar.It is because there are too many warnings…… And I found this,it looks like the input image is too small when using tta
> 
> ```
> RuntimeError: Calculated padded input size per channel: (1 x 6). Kernel size: (2 x 2). Kernel size can't be greater than actual input size
> ```
I haven't encountered this problem, perhaps you can try increasing the zoom ratio in tta_pipeline to avoid the image being too small.

xiexinch:
Hi @ZiAn-Su,
Would you update this PR? If you're busy, I can try to fix it.

ZiAn-Su:
> Hi @ZiAn-Su, Would you update this PR? If you're busy, I can try to fix it.
Of course, sorry to be not familiar with the submission process. I'll update it immediately.

ZiAn-Su:
> > Hi @ZiAn-Su, Would you update this PR? If you're busy, I can try to fix it.
> > Of course, sorry to be not familiar with the submission process. I'll update it immediately.

The new PR has been submitted. 😀

xiexinch (approved):

Issue #2808, opened 2023-03-29
Add BDD100K Dataset


SnowdenLee:
# Describe the feature

**Motivation**
Integrate [BDD100K](https://paperswithcode.com/dataset/bdd100k) dataset. It shares the same classes as Cityscapes, and it's commonly used for evaluating segmentation/detection tasks in driving scenes, such as in [RobustNet](https://arxiv.org/abs/2103.15597), [WildNet](https://github.com/suhyeonlee/WildNet).

**Related resources**
[Here](https://github.com/shachoi/RobustNet/blob/main/datasets/bdd100k.py) is an example for loading BDD100K dataset.

PR #3158, opened 2023-07-02
[Feature] add bdd100K datasets

CastleDream:
## Motivation
Integrate [BDD100K](https://paperswithcode.com/dataset/bdd100k) dataset. It shares the same classes as Cityscapes, and it's commonly used for evaluating segmentation/detection tasks in driving scenes, such as in [RobustNet](https://arxiv.org/abs/2103.15597), [WildNet](https://github.com/suhyeonlee/WildNet).

Enhancement for Add BDD100K Dataset #2808


mmseg/datasets/bdd100k.py
```diff
@@ -0,0 +1,31 @@
...
+                 'traffic light', 'traffic sign', 'vegetation', 'terrain',
+                 'sky', 'person', 'rider', 'car', 'truck', 'bus', 'train',
+                 'motorcycle', 'bicycle', 'Unlabeled'),
+        palette=[[128, 64, 128], [244, 35, 232], [70, 70, 70], [102, 102, 156],
+                 [190, 153, 153], [153, 153, 153], [250, 170,
+                                                    30], [220, 220, 0],
+                 [107, 142, 35], [152, 251, 152], [70, 130, 180],
+                 [220, 20, 60], [255, 0, 0], [0, 0, 142], [0, 0, 70],
+                 [0, 60, 100], [0, 80, 100], [0, 0, 230], [119, 11, 32],
+                 [0, 0, 0]])
```
xiexinch, line 20: Could we remove the `Unlabeled` class?

xiexinch (approved):

mmseg/datasets/bdd100k.py
CastleDream: Yes，we should keep consistent with Cityscapes.  In [bdd100k/label.py]
(https://github.com/bdd100k/bdd100k/blob/master/bdd100k/label/label.py#L42C47-L42C47) mentions:
```python
# Our extended list of label types. Our train id is compatible with Cityscapes
labels = [
    #       name                     id    trainId   category catId
    #       hasInstances   ignoreInEval   color
    Label("unlabeled", 0, 255, "void", 0, False, True, (0, 0, 0)),
```
but i am confused that why some dataset.py, for example in [datasets/mapillary.py](https://github.com/open-mmlab/mmsegmentation/blob/f52bd5e456415dd707179eaed2e25a06091b8dc8/mmseg/datasets/mapillary.py), including `Unlabeled`, and most other dataset.py, didn't include.

and in the openMMLabCamp classes video, the teacher also add `Unlabeled` when import custom dataset..

xiexinch: This might depend on the requirements at the time of creating the dataset, and if there is a need to accurately recognize the background class, then it is necessary to include this class.

xiexinch (approved):

mmseg/datasets/bdd100k.py
CastleDream: Fine, thanks for your apply!

PR #3235, opened 2023-07-27
[Fix] Fix module PascalContextDataset

angiecao:
## Motivation

-  'PascalContextDataset' object has no attribute 'file_client', it will cause an error. 
-  The attribute ‘ann_file’ is not allowed to be empty, otherwise, an error will be reported.

## Modification
- Replace file_client with fileio

### DIFF
--- a/mmseg/datasets/pascal_context.py
+++ b/mmseg/datasets/pascal_context.py
@@ -10,8 +10,8 @@ class PascalContextDataset(BaseSegDataset):
     """PascalContext dataset.
 
     In segmentation map annotation for PascalContext, 0 stands for background,
-    which is included in 60 categories. ``reduce_zero_label`` is set to True.
-    The ``img_suffix`` is fixed to '.jpg' and ``seg_map_suffix`` is
+    which is included in 60 categories. ``reduce_zero_label`` is fixed to
+    False. The ``img_suffix`` is fixed to '.jpg' and ``seg_map_suffix`` is
     fixed to '.png'.
 
     Args:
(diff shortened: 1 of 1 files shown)


mmseg/datasets/pascal_context.py
```diff
@@ -46,18 +46,18 @@ class PascalContextDataset(BaseSegDataset):
                  [255, 71, 0], [0, 235, 255], [0, 173, 255], [31, 0, 255]])
 
     def __init__(self,
-                 ann_file: str,
+                 ann_file='',
                  img_suffix='.jpg',
                  seg_map_suffix='.png',
+                 reduce_zero_label=True,
```
xiexinch, line 52: We can't set `reduce_zero_label` to `True` here.
https://mmsegmentation.readthedocs.io/en/latest/notes/faq.html#functionality-of-reduce-zero-label

mmseg/datasets/pascal_context.py
```diff
@@ -100,16 +100,15 @@ class PascalContextDataset59(BaseSegDataset):
                  [255, 71, 0], [0, 235, 255], [0, 173, 255], [31, 0, 255]])
 
     def __init__(self,
-                 ann_file: str,
+                 ann_file='',
                  img_suffix='.jpg',
                  seg_map_suffix='.png',
-                 reduce_zero_label=True,
+                 reduce_zero_label=False,
```
xiexinch, line 106: And here it should be `True`.

xiexinch (approved):

mmseg/datasets/pascal_context.py
angiecao: Thank you for your correction. I think I was confused because when I created the annotation maps, I had set the background to 255, and the category labels are 0-59, which caused me to set recude_zero_label to False in class ''PascalContextDataset59''. I've restored the settings for reduce_zero_label.

PR #3348, opened 2023-09-23
[Fix] fix wrong variables passing for `set_dataset_meta`

likyoo:
see [here](https://github.com/open-mmlab/mmsegmentation/blob/dev-1.x/mmseg/visualization/local_visualizer.py#L223C13-L223C13)


xiexinch (approved):

xiexinch:
Thanks!

Issue #3461, opened 2023-12-03
Feature request: Support ACDC Dataset


CynthiaLiNingxin:
# Describe the feature

**Motivation**
There is a recent paper ACDC: The Adverse Conditions Dataset with Correspondences for Semantic Driving Scene Understanding, which is very helpful for semantic scene understanding in adverse weather.

**Related resources**
ACDC official website: https://acdc.vision.ee.ethz.ch/

**Additional context**
It would be great if mmsegmentation could support this dataset.

Tonyzhang2000:
Hey thank you for the request, I would like to take this issue.

PR #3470, opened 2023-12-10
[Feature] Support ACDC dataset

tonyzstt:
## Motivation

As requested by #3461

## Modification

- Add ACDC data loader and config file
- Add ACDC data processing script
- Update documentation

### DIFF
--- a/tools/dataset_converters/acdc.py
+++ b/tools/dataset_converters/acdc.py
@@ -24,12 +24,6 @@ def parse_args():
 def main():
     args = parse_args()
 
-    try:
-        shutil.rmtree(args.out_dir)
-        print('Directory removed successfully')
-    except OSError as e:
-        print(f'Error: {e.strerror}')
-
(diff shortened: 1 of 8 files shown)


tools/dataset_converters/acdc.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def main():
+    args = parse_args()
+
+    try:
+        shutil.rmtree(args.out_dir)
+        print('Directory removed successfully')
+    except OSError as e:
+        print(f'Error: {e.strerror}')
```
xiexinch, line 31: This is a dangerous operation and it is best not to delete any of the user's files.

Tonyzhang2000: Updated!

Tonyzhang2000:
Hi @xiexinch, I'm wondering if anything else needs to be changed. If not, can this PR be merged?

Issue #3427, opened 2023-11-09
how to visualize the result without background?


shinpaul14:
Hello. 

I just had a question about visualization. 

I am hoping to overpay my model's prediction to the input image, however, the appearance of the background and the name tag is the big problem. 
How can I remove them when I visualize my results?

Thank you.

guozhen-dev:
According to my understanding, I assume you want to have an export result image without a background and the label text. I am working on supporting a switch to enable the selection of layers on this.

guozhen-dev:
Just want to make sure this is the feature you want to have. 
<img width="1028" alt="image" src="https://github.com/open-mmlab/mmsegmentation/assets/39869385/0e8f51fd-83ac-484f-ae21-42bddecd0439">

shinpaul14:
Yes, I want to show only the prediction without the background and the name tag.

xiexinch:
Please check the use case in this PR https://github.com/open-mmlab/mmsegmentation/pull/3466

bjzhb666:
@xiexinch @guozhen-dev  Hi, I am running tools/test.py and facing the same problem. How to solve it?
 https://github.com/open-mmlab/mmsegmentation/issues/3648

```
python tools/test.py configs.py \
 work_dirs/0.pth \
 --show-dir work_dirs
```

PR #3473, opened 2023-12-12
Enable the segement_only flag

Tonyzhang2000:
## Motivation

This PR provided a simple flag that the user can opt out of all the other information in the rendered image, and only keep the segmentation information. Resolve issue https://github.com/open-mmlab/mmsegmentation/issues/3427.

## Modification

A new flag segement_only (Optional) is added to the SegLocalVisualizer class, enabling users to only display the segmentation information if set to true.
image

### DIFF
--- a/demo/image_demo.py
+++ b/demo/image_demo.py
@@ -21,6 +21,10 @@ def main():
         help='Opacity of painted segmentation map. In (0, 1] range.')
     parser.add_argument(
         '--title', default='result', help='The image identifier.')
+    parser.add_argument(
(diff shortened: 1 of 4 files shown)


xiexinch:
Hi @Tonyzhang2000,
Thanks for your contribution, but this modification is applied with the PR https://github.com/open-mmlab/mmsegmentation/pull/3466. You might check the `use case` at the PR. If there are other bugs, please raise a new issue or PR to notice us, we are really appriciate it!

PR #3551, opened 2024-01-31
[Fix] Fix get_flops scrip

xiexinch:
## Motivation

`inputs` and `input_shape` can't be both set to mmengine api `get_model_complexity_info`

https://github.com/open-mmlab/mmengine/pull/1056

## Modification

Set `input_shape` to None.


Deom-z:
请问还是不支持maskformer和mask2former吗

PR #3649, opened 2024-04-23
[Fix] Fix testing without labels features

bjzhb666:

## Motivation

Now, testing with labels is the default setting, which is not common.

## Modification

Change the default value from True to False

## BC-breaking (Optional)

NO.

## Use cases (Optional)

No need.

## Checklist


MGAMZ (commented):
Maybe it's a better solution to modify this param in config, rather than directly modify default values globally, considering most datasets have named labels.

PR #3659, opened 2024-05-06
[Dev] Update mmcv max version

xiexinch:
Update mmcv version in requirements.
https://github.com/open-mmlab/mmcv/releases/tag/v2.2.0


mmseg/__init__.py
```diff
@@ -8,7 +8,7 @@
 from .version import __version__, version_info
 
 MMCV_MIN = '2.0.0rc4'
-MMCV_MAX = '2.2.0'
+MMCV_MAX = '3.0.0'
```
MGAMZ, line 11: MMCV3.0 does not exist now, please refer to pr#3663

MGAMZ (changes requested):
rude override for version is not suggested.

mmcv3.0 has not been officially released so far.

PR #3675, opened 2024-05-20
[Fix] bug fix for old MMCV format for mobilenetv3 configs file (changed to MMEngine format)

tackhwa:
the configs file for mobilenetv3 is using "schedule_160k.py" , which is not fulfill training setting in [this link](https://github.com/tackhwa/mmsegmentation/blob/main/configs/mobilenet_v3/README.md). It should be 320k iterations instead of 160k. and the "Runner" is deleted as it is old MMCV format configs.

- Before: 
pasted image 0

- After: 
pasted image 0


tackhwa:
@xiexinch could you check this pr, please。

Issue #3666, opened 2024-05-11
Inference of mask2former on large images


mboaz17:
Hello,

I am trying to do train and evaluate a mask2former model on large images (3648x5472 pixels).
Training is not an issue in terms of memory usage, since I'm using cropped patches - the train_pipeline contains random cropping:
    dict(type='RandomCrop', crop_size=crop_size, cat_max_ratio=0.75)

However, inference of the model on such large images requires huge memory.
For example, for the mask2former_swin-b model, about 40GB of GPU memory are necessary.

This is why I tried changing the test_cfg of the model from 
    test_cfg=dict(mode='whole'))
to sliding window inference:
    test_cfg=dict(mode='slide', crop_size=(1366, 2048), stride=(1141, 1712)))

However, during the evaluation phase of training I received the following error:

  File "/home/airsim/repos/open-mmlab/mmsegmentation/tools/train.py", line 100, in main
    runner.train()
  File "/home/airsim/anaconda3/envs/open-mmlab-newest/lib/python3.8/site-packages/mmengine/runner/runner.py", line 1777, in train
    model = self.train_loop.run()  # type: ignore
  File "/home/airsim/anaconda3/envs/open-mmlab-newest/lib/python3.8/site-packages/mmengine/runner/loops.py", line 102, in run
    self.runner.val_loop.run()
  File "/home/airsim/anaconda3/envs/open-mmlab-newest/lib/python3.8/site-packages/mmengine/runner/loops.py", line 371, in run
    self.run_iter(idx, data_batch)
  File "/home/airsim/anaconda3/envs/open-mmlab-newest/lib/python3.8/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/airsim/anaconda3/envs/open-mmlab-newest/lib/python3.8/site-packages/mmengine/runner/loops.py", line 391, in run_iter
    outputs = self.runner.model.val_step(data_batch)
  File "/home/airsim/anaconda3/envs/open-mmlab-newest/lib/python3.8/site-packages/mmengine/model/base_model/base_model.py", line 133, in val_step
    return self._run_forward(data, mode='predict')  # type: ignore
  File "/home/airsim/anaconda3/envs/open-mmlab-newest/lib/python3.8/site-packages/mmengine/model/base_model/base_model.py", line 361, in _run_forward
    results = self(**data, mode=mode)
[...]

PR #3752, opened 2024-08-05
[Fix] Fix dimension error when using slide inference with Mask2Former head

Joris-Kuehl-TU-Berlin:
## Motivation

The method `predict` in [mmseg/models/decode_heads/mask2former_head.py](https://github.com/open-mmlab/mmsegmentation/blob/b040e147adfa027bbc071b624bedf0ae84dfc922/mmseg/models/decode_heads/mask2former_head.py#L153) is incompatible with the `'slide'` inference mode.

To elaborate, on line 280 of `slide_inference` in [mmseg/models/segmentors/encoder_decoder.py](https://github.com/open-mmlab/mmsegmentation/blob/b040e147adfa027bbc071b624bedf0ae84dfc922/mmseg/models/segmentors/encoder_decoder.py#L280), the key `'img_shape'` of `batch_img_metas[0]` is overwritten by the shape of the cropped image / sliding window.

In line 353 of [mmseg/models/decode_heads/decode_head.py](https://github.com/open-mmlab/mmsegmentation/blob/b040e147adfa027bbc071b624bedf0ae84dfc922/mmseg/models/decode_heads/decode_head.py#L353), this is the first key that is used to get the target size for upsampling. As such, `crop_seg_logits` will have the same shape as `crop_img`.

In [mmseg/models/decode_heads/mask2former_head.py](https://github.com/open-mmlab/mmsegmentation/blob/b040e147adfa027bbc071b624bedf0ae84dfc922/mmseg/models/decode_heads/mask2former_head.py#L153), the first shape that is referenced as a target for upsampling is `'pad_shape'` instead. Since `slide_inference` does not overwrite this key, each cropped image is upsampled to the size of the full image, and then further padded with zeros by `slide_inference`, leading to a dimension mismatch when trying to add up the padded `crop_seg_logits`. This leads to the issue described in #3666.

## Modification

I have simply adjusted the size selection in [mmseg/models/decode_heads/mask2former_head.py](https://github.com/open-mmlab/mmsegmentation/blob/b040e147adfa027bbc071b624bedf0ae84dfc922/mmseg/models/decode_heads/mask2former_head.py#L153) to match that of [mmseg/models/decode_heads/decode_head.py](https://github.com/open-mmlab/mmsegmentation/blob/b040e147adfa027bbc071b624bedf0ae84dfc922/mmseg/models/decode_heads/decode_head.py#L353). As a result, a dimension mismatch no longer occurs when using slide inference with a Mask2Former head.

## Checklist

### DIFF
--- a/mmseg/models/decode_heads/mask2former_head.py
+++ b/mmseg/models/decode_heads/mask2former_head.py
@@ -150,8 +150,11 @@ class Mask2FormerHead(MMDET_Mask2FormerHead):
         all_cls_scores, all_mask_preds = self(x, batch_data_samples)
         mask_cls_results = all_cls_scores[-1]
         mask_pred_results = all_mask_preds[-1]
-        if 'pad_shape' in batch_img_metas[0]:
-            size = batch_img_metas[0]['pad_shape']
+        if isinstance(batch_img_metas[0]['img_shape'], torch.Size):
+            # slide inference
+            size = batch_img_metas[0]['img_shape']
+        elif 'pad_shape' in batch_img_metas[0]:
+            size = batch_img_metas[0]['pad_shape'][:2]
         else:
             size = batch_img_metas[0]['img_shape']
         # upsample mask


londumas:
Thank you for this PR, it works.

PR #3755, opened 2024-08-08
Update confusion_matrix.py

MGAMZ:
## Motivation

The current implementation do not work correctly when num_classes is larger than 16. Because the gt seg map was forced to uint8 dtype.

## Modification

Let gt seg map being np.uint32, rather than uint8 for no reason. The function works fine after the modification.


MGAMZ:
This repo is lack of maintainance, so I'm closing this PR.

